PR #64, opened 2014-09-25
preserve existing 'regressions' options when parsing command line opts

tmcdonell:
When using 'defaultMainWith', the existing command line parser would discard the existing values for 'Config.regressions' and replace it with only the new "--regress=..." entries it parses. This change preserves the old values as well.

### DIFF
--- a/Criterion/Main/Options.hs
+++ b/Criterion/Main/Options.hs
@@ -111,9 +111,9 @@ config Config{..} = Config
   <*> option (range 1 1000000)
       (long "resamples" <> metavar "COUNT" <> value resamples <>
        help "number of bootstrap resamples to perform")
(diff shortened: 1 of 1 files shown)


bos:
The current behaviour is deliberate, and I'm not thrilled about this change to only allow addition of regressions to a default set – especially as if someone depends on the current behaviour, they'll be surprised.

tmcdonell:
I don't understand what you mean by this "only allowing additions to the default set".

Anyway, if that's the intended behaviour, fair enough. I think that you should at least document that `defaultMainWith` ignores this part of the `Config` structure it is passed --- that is surprising and non-intuitive behaviour, which is the crux of the issue.

PR #67, opened 2014-10-21
Replace EitherT with ExceptT from mtl, drop dependency on either package...

xich:
As far as I can tell, the `ExceptT` transformer in `mtl` does the same thing `EitherT` does (modulo not having a nice name for the hoist function).

The motivation here is that a dependency on the `either` package brings in a whole load of transitive deps, whereas we already have `mtl`. This patch lowers total dependencies from 54 to 37.


bos:
This worries me because it seems likely to cause massive compatibility problems with older versions of mtl.

xich:
Yeah, it occurs to me now that one does not simply bump the lower bounds on mtl on popular packages like criterion. :-P

Though, there's a package for this...

http://hackage.haskell.org/package/transformers-compat

Essentially, you just add a dependency on transformers-compat, and it adds the `ExceptT` type when an older `mtl` and `transformers` is installed, or doesn't add it when a new enough one (that has `ExceptT` already) is present.

xich:
I'll try to find a 7.6 install and give it a shot and update my patch.

bos:
I'm giving that idea a try with 4a57ab10a2b56aebcc5a060ddfcb55c4903cbb50.

xich:
Turns out `transformers-compat` is already a transitive dep of criterion with the latest install set and mtl-2.1.2, which is the only old version I had handy to try. It builds anyway, haven't tried running the test suite.

PR #72, opened 2015-01-04
Use the js-jquery and js-flot libraries rather than copying all the source

ndmitchell:
Criterion uses the Flot and jQuery Javascript libraries. At the moment, it keeps copies locally and also has copies of the source code to be compliant with Debian conditions. It doesn't actually ship the licenses for those libraries, so it might not technically be in compliance even now.

This patch switches to the js-jquery and js-flot libraries on Hackage, which are dead simple, and just contain the jquery/flot code, including license and sources to make Debian happy. The Shake library already uses them, and I'm hoping to make them the "standard" way of getting this code.

As a side effect, I've also switched to the 1.* branch of jQuery rather than 2.*. The 2.* branch has no additional features or performance, but has worse compatibility, so currently 1.* is recommended. If that's a conscious decision, I'm happy to build a version of the js-jquery library for 2.*.


rrnewton:
@RyanGLScott looked this over and agrees with it.

rrnewton:
Curious -- it looks like accidentally closing and reopening gives travis a kick to get going.

RyanGlScott:
Yep, I'd consider that a feature. Sometimes, a Travis CI build fails because you didn't wait long enough to wait for `cabal update` to get the latest stuff. If you don't have permission to restart the build via Travis, you can do it yourself by closing and re-opening the PR :)

PR #73, opened 2015-01-21
add `--match pattern` and `--match ipattern` matchers

osa1:
Interestingly, I found prefix matcher completely useless and glob matcher hard to use(can that even handle case insensitivity? I don't think so).

I guess most general solution would be to provide a regex matcher, but to keep it simple I wanted to go with these simple solutions.

### DIFF
--- a/Criterion/Main.hs
+++ b/Criterion/Main.hs
@@ -53,7 +53,8 @@ import Criterion.Main.Options (MatchType(..), Mode(..), defaultConfig, describe,
 import Criterion.Measurement (initializeTime)
 import Criterion.Monad (withConfig)
 import Criterion.Types
-import Data.List (isPrefixOf, sort, stripPrefix)
+import Data.Char (toLower)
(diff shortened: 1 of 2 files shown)


rrnewton:
Should this be called "pattern" or "substr"?  It does exact matching only, right?

mgsloan:
This'd be a welcome addition

rrnewton:
This has tripped me up also.  Can anyone actually get the glob syntax to work?  I've always put in the full name of the benchmark.

mgsloan:
Yeah, I've never gotten criterion's benchmark selection stuff to work.  I just comment out the stuff I don't want to run...

rrnewton:
We discussed this a bit today and `pattern` doesn't seem that descriptive to me.  I'm thinking of `infix` to match the existing `prefix`.

rrnewton:
Btw, as you can see in that commit above.  I resolved the conflicts, but this needs one more tweak to update the help messages.  Continuing this on a new PR, #123.

PR #74, opened 2015-02-12
New feature: algorithm comparison graphs

k32:
Hello,

First of all, thanks for your fantastic library. It has proven itself very useful for choosing an optimal implementation. Speaking of which, I tried to make results a bit more vivid by introducing performance comparison graphs like this:
 
vsgraph

It can be done using a new API (bversus is like a special kind of groups). Simple example:

```haskell
main = defaultMain [bgroup "foo" [
                       bversus "sum" [(i,) `fmap` replicateM i $
                                        randomRIO (0::Int, 100)
                                       | i<-[100, 200, 400, 800, 1600]]
                         [
                           ("prelude", nf sum)
                         , ("foldl'", nf $ foldl' (+) 0)
                         , ("foldr", nf $ foldr (+) 0)
                         ]
                       ]
                   ]
```

I have to admit that the quality of code may be not very good and there is a room for improvement. Moreover, I haven't touched JS before, so the quality of the JS code is probably horrific.

Anyway, I would be grateful if you look through this feature and say if it worth further development.

### DIFF
--- a/Criterion.hs
+++ b/Criterion.hs
@@ -59,4 +59,4 @@ benchmarkWith' cfg bm = do
   initializeTime
   withConfig cfg $ do
     _ <- note "benchmarking...\n"
-    runAndAnalyseOne 0 (ROBench "function") bm
+    runAndAnalyseOne 0 [ROBench "function"] bm
--- a/Criterion.hs
+++ b/Criterion.hs
@@ -19,6 +19,7 @@ module Criterion
     , env
     , bench
     , bgroup
+    , bversus
     -- ** Running a benchmark
     , nf
     , whnf
--- a/Criterion.hs
+++ b/Criterion.hs
@@ -59,4 +59,4 @@ benchmarkWith' cfg bm = do
   initializeTime
   withConfig cfg $ do
     _ <- note "benchmarking...\n"
-    runAndAnalyseOne 0 "function" bm
+    runAndAnalyseOne 0 (ROBench "function") bm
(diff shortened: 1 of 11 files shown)


bos:
Hi, @k32 – thank you for doing all of this work.

First of all, I like the idea behind this, and I would like an evolution of your code to go in. However, I need your help. Your changes as they currently stand are not easy to review, as they "show your work" too much. I do not want  to work through intermediate commits like "a bit less ugly".

Here's what I would like to see: a series of small self-contained commits, each doing just one thing. For example, you've changed the type of `writeCsv` in some commit along with a lot of other things, but this should really be a single commit that does nothing else, and that compiles and passes the tests – and all of the other commits should be like this too. Please feed these changes to me in smaller increments, so that I can give you feedback on each commit and you won't have to change follow-on commits in response to comments on earlier commits in the series.

I would also like you to more closely follow the existing coding style of the project, so that a reader will not be able to tell the difference between code you wrote and code I wrote.

Finally, please no cryptic function names like `vscsv`. This stuff has to be easy for me to read and maintain, as the burden of doing so will fall to me.

By the way, these are all "good neighbour" suggestions for contributing to any open source project. I realise I'm asking you to do several more hours of work to clean up your effort so that I can review it, but it will make review and subsequent maintenance much easier and safer for me.

k32:
Hi,

I agree with your comments, I hope I will have time to apply the remarks this week. Also I was thinking of better reusing of the exiting code. For example, I could make better usage of benchmark environments instead of reimplementing them.

rrnewton:
We're triaging old issues and PRs a bit -- @k32, would you still like to push ahead with this?

k32:
Hello,

It's been a while since the first attempt. I'm afraid that the baseline code has changed significantly, and if I'll really decide to push it forward, the commit should be reworked completely.

So it's better to close this pull request. A new one will be created if needed.

PR #89, opened 2015-10-02
Flush stdout and stderr so messages appear immediately even when piped

Rufflewind:
This ensures that messages will appear in a timely manner even when piped through a program like `tee`.


osa1:
:+1: this is a very annoying issue when running big number of benchmarks. (you have to `tee` otherwise some benchmarks get lost in the terminal history etc.)

bos:
Nice, thanks!

PR #93, opened 2016-01-07
Improve help and error message of --match: Show values alternatives

osa1:
Prior to this patch it was basically impossible, without looking at the source code.


bos:
Nice, thanks!

PR #94, opened 2016-01-07
Criterion.Main: Factor out and export runMode

bgamari:
Previously Criterion.Main.Options exported the Mode type which is great
when you want to add a few domain-specific options to your benchmark.
Unfortunately, there was no exported function to actually run the
benchmark as defaultMain would have done. This adds precisely this
function.


osa1:
This is actually one of the things I was planning to make.. :+1: thanks @bgamari .

bos:
Nice, thanks!

PR #98, opened 2016-02-25
cbits/cycles.c: Add support for Linux on non-Intel CPUs

erikd:
Tested on armhf and powerpc.


bos:
Nice, thanks!

erikd:
Thanks @bos. Now I'll add criterion to my personal Jenkins instance and make sure it gets built on armhf or powerpc on a regular basis.

PR #103, opened 2016-06-08
Add JSON Report serialization as an alternative to `--raw`

rrnewton:
We've been using this in our benchmarking and it completely removes our dependence on the (buggy) `--raw` output.

@RyanGLScott should be able to confirm that all 1800 stackage benchmarks run while using JSON serialization internally.  (That is, during the benchmarking loop, reports are written out and read back in in json format.)  This fixes a number of failing cases within those stackage benchmarks.

### DIFF
--- a/Criterion/Internal.hs
+++ b/Criterion/Internal.hs
@@ -22,7 +22,7 @@ import qualified Data.Aeson as Aeson
 import Control.Applicative ((<$>))
 import Control.DeepSeq (rnf)
 import Control.Exception (evaluate)
-import Control.Monad (foldM, forM_, void, when)
+import Control.Monad (foldM, forM_, void, when, unless)
 import Control.Monad.Reader (ask, asks)
 import Control.Monad.Trans (MonadIO, liftIO)
 import Control.Monad.Trans.Except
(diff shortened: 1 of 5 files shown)


Criterion/Internal.hs
```diff
@@ -111,33 +112,45 @@ runAndAnalyse :: (String -> Bool) -- ^ A predicate that chooses
...
 runAndAnalyse select bs = do
-  mbRawFile <- asks rawDataFile
-  (rawFile, handle) <- liftIO $
-    case mbRawFile of
+  mbJsonFile <- asks jsonFile
+  (jsonFile, handle) <- liftIO $
+    case mbJsonFile of
       Nothing -> do
         tmpDir <- getTemporaryDirectory
-        openBinaryTempFile tmpDir "criterion.dat"
```
RyanGlScott: IIUC, `criterion.dat` is used for incrementally writing out the binary report. Are we breaking `--raw` further by removing it?

rrnewton:
Leaving this here for documentation.  The recent update to master already has these changes.

Criterion/Internal.hs
rrnewton: Oops!  Good catch.  Specifically, "criterion.dat" should only be used  when `rawDataFile` is *not* set.

But you are right that we forgot to restore writing of binary reports when `rawDataFile` *is* set.  I.e. just below where it calls `report rpts; json rpts; junit rpts`.

rrnewton: Ok, fixed here: https://github.com/bos/criterion/commit/92d6279ca24aead14351d38e5157bfaba3153c1d

PR #104, opened 2016-06-08
Restore compat with GHC 7.4

rrnewton:
This is just an update of https://github.com/bos/criterion/pull/99

It resolves a merge conflict.

CC @hvr

### DIFF
--- a/criterion.cabal
+++ b/criterion.cabal
@@ -112,7 +112,6 @@ library
 --   main-is:        App.hs
 --   other-modules:
 --     Options
---     Paths_criterion
 
 --   ghc-options:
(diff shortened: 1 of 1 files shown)


rrnewton:
Related to this -- if we want to actually *test* the GHC 7.4 compat on travis that would be another matter.  Currently we've gone with a stack-only setup in the `.travis` file and there is no LTS that covers 7.4.

rrnewton:
Oops, belatedly realized that this only touches the executable, which is now commented.  So it becomes trivial.  It doesn't hurt though -- just in case we reactivate that executable.

Issue #107, opened 2016-06-09
ERROR: error reading file sanity.json


erikd:
After @ndmitchell 's patch to remove the embedded `js-jquery` and `js-flot` libraries, the tests fail for me (on one of two machines) with:

```
$ dist/build/sanity/sanity
benchmarking fib/fib 10
time                 2.858 μs   (2.854 μs .. 2.863 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 2.861 μs   (2.856 μs .. 2.867 μs)
std dev              17.83 ns   (13.84 ns .. 23.61 ns)

benchmarking fib/fib 22
time                 2.034 ms   (2.028 ms .. 2.040 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 2.050 ms   (2.045 ms .. 2.060 ms)
std dev              23.57 μs   (17.58 μs .. 35.98 μs)

benchmarking length . filter/string
time                 3.206 μs   (3.203 μs .. 3.210 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 3.207 μs   (3.204 μs .. 3.211 μs)
std dev              12.23 ns   (9.303 ns .. 16.52 ns)

benchmarking length . filter/bytestring
time                 892.5 ns   (890.2 ns .. 896.1 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 897.3 ns   (894.3 ns .. 902.2 ns)
std dev              13.40 ns   (7.963 ns .. 22.55 ns)
variance introduced by outliers: 15% (moderately inflated)

sanity: [Failed]
ERROR: error reading file sanity.json:
  "Error in $: endOfInput"
```

The two machines are pretty much identical, running Debian Testing on x86_64.

ndmitchell:
When you say pretty much identical, are there any Haskell library version differences between the two?  That would be a good hint at what might have gone wrong.

erikd:
I did a `cabal update` and rebuilt on the machine that was working and now it fails like the other machine.

ndmitchell:
I can't reproduce. @erikd - can you try using something like https://www.stackage.org/nightly-2016-06-10/cabal.config which pins down absolutely all version ranges and see what happens? The Travis test suite seemingly passed with stack nightly.

ndmitchell:
I just did a completely fresh Windows VM, GHC 8.0.1, update, test and it worked perfectly.

[...]

PR #108, opened 2016-06-11
#107, open the output file in the write mode

ndmitchell:
With ReadWriteMode Haskell opens the file but doesn't truncate the contents. That means if we have a file there from before, we end up overwriting the start, and if the file was longer, we end up with corrupt JSON. Only visible if you run the test suite multiple times, and somewhat non-deterministic since it requires longer times (when printed in exponential format), followed by shorter times.


rrnewton:
Thanks for catching this.  I fee silly for reproducing that bug because I think something similar may have been happening with the previous binary serialization.

PR #109, opened 2016-06-13
Fix link to examples in README.markdown

zyv:
GitHub link processing magic takes care of adding the right repo & branch.


rrnewton:
Thanks. Nice trick.

PR #112, opened 2016-08-12
Implement get for Binary Regression instance

RyanGlScott:
For some reason, it's not currently implemented, and it's resulted in [at least one sporadic failure](https://travis-ci.org/RyanGlScott/text-show/jobs/151838151#L1835) for me during benchmarking. I'm not sure _why_ it's sporadic, but luckily it's an easy fix.


RyanGlScott:
This seems to happen more reliably when `criterion` is build with GHC 7.6, since it ships with an old version of `binary` which doesn't include a `DefaultSignatures`-based default implementation for `get`.

Issue #79, opened 2015-03-21
`env` like function for freeing resources


osa1:
I have some number of benchmarks and in each one of them, I'm allocating some resources from a limited pool of resources. I need to free the resources after each benchmark, because rest of the benchmarks will be using those resources, however, I don't want to measure time spend on freeing the resources. As far as I can see Criterion doesn't have a way to do this right now. Implementing this would be really awesome.

mgsloan:
+1

I just needed this as well.

mgsloan:
It would also be nice to have a variant which doesn't require an `NFData` instance for the environment, as it could be the responsibility of the environment initialization code to do such evaluation.

osa1:
> It would also be nice to have a variant which doesn't require an NFData instance for the environment, as it could be the responsibility of the environment initialization code to do such evaluation.

Can't you just wrap your data with a newtype, and implement `instance NFData YourData where rnf (YourData _) = ()` ?

mgsloan:
True!  Maybe the `NFData` constraint isn't so bad to let the user know they should consider this.

PR #124, opened 2016-09-09
Add `envWithCleanup`

tvh:
This is a fix for #79

### DIFF
--- a/Criterion/Internal.hs
+++ b/Criterion/Internal.hs
@@ -182,13 +182,15 @@ for :: MonadIO m => (String -> Bool) -> Benchmark
     -> (Int -> String -> Benchmarkable -> m ()) -> m ()
 for select bs0 handle = go (0::Int) ("", bs0) >> return ()
   where
-    go !idx (pfx, Environment mkenv mkbench)
+    go !idx (pfx, Environment mkenv cleanupenv mkbench)
       | shouldRun pfx mkbench = do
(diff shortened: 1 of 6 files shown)


Criterion/Internal.hs
```diff
@@ -182,13 +182,15 @@ for :: MonadIO m => (String -> Bool) -> Benchmark
...
-    go !idx (pfx, Environment mkenv mkbench)
+    go !idx (pfx, Environment mkenv cleanupenv mkbench)
       | shouldRun pfx mkbench = do
         e <- liftIO $ do
           ee <- mkenv
           evaluate (rnf ee)
           return ee
-        go idx (pfx, mkbench e)
+        idx' <- go idx (pfx, mkbench e)
+        liftIO $ cleanupenv e
```
mgsloan: I don't think that is quite correct.  You want something like `finally (go idx (pfx, mkbench e)) (liftIO (cleanupenv e))`

mgsloan:
Thanks for working on this!  This isn't correct for exceptions.  The exception case is the tricky one, and the entire reason an explicit cleanup action is required.

A test should also be added to make sure it works with exceptions.

tvh:
Sorry for taking this long to get back at this.

I know this is nor correct for exceptions, but I assumed it is okay, as

RyanGlScott:
This has been superseded by #136.

Issue #114, opened 2016-08-16
Report measurements for --iters (and fix parsing errors)


rrnewton:
@RyanGlScott is looking into command parsing errors that prevent `--iters` to be used with other command line options.

Further, it would be great to report SOME information from such a run.  Criterion cannot do a regression with just one batch of iters, but it can at least report the time.

RyanGlScott:
Oh, well finding out the source of the parser "error" turned out to be [surprisingly easy](https://github.com/bos/criterion/blob/32afe61935e55662425d535df2edbfd30bb79afd/Criterion/Main/Options.hs#L84-L93):

```haskell
parseWith :: Config -> Parser Mode
parseWith cfg =
    (matchNames (Run <$> config cfg)) <|>
    runIters <|>
    (List <$ switch (long "list" <> short 'l' <> help "List benchmarks")) <|>
    (Version <$ switch (long "version" <> help "Show version info"))
```

It turns out that `--iters` is an entirely different execution mode than the mode that allows specifying other options.

How easy it is to fold `--iters` into the other options, I don't know...

merijn:
As far as I can tell the fairly trivial change in my PR #136 should also fix this.

Issue #135, opened 2017-04-25
Benchmarking IO operations that grow the amount of live data.


merijn:
Criterion runs many batches of many repeats of the same benchmark to acquire timings. This is fine for pure functions, but when benchmarking IO operations that grow the heap this can lead to earlier runs impacting the timings of later runs. Since GHC's copying GC results in longer GC pauses the bigger the amount of live data.

For example: Suppose we want to benchmark writing to a `Chan`. Naively we can use `env` to allocate a `Chan` for our benchmark and benchmark `writeChan env ()` to time writing to the `Chan`. However, since the entire `Chan` remains live over many runs of this action we end up repeatedly growing it's contents, resulting in later runs being disproportionately affected by GC work from earlier runs. If criterion runs 1,000 batches of 1,000 repeats, the last batch has to GC 999,000 values in the `Chan` unrelated to that benchmark.

Ideally Criterion would allow benchmarks that share an environment only over 1 one "batch" of benchmarks, so that in our previous example we allocate one `Chan`, perform 1,000 writes, then GC it and allocate a fresh `Chan` for the next batch of 1,000 writes. This should perform more realistic timings.

RyanGlScott:
Fixed in #136.

Issue #134, opened 2017-04-25
Benchmarking mutable IO operations


merijn:
In it's current form Criterion benchmarks that need an environment always share the environment over many runs of the same benchmark. This is fine for timing simple IO operations and pure functions, but makes it hard to benchmark mutable operations such a sorting a mutable Vector in place.

Ideally Criterion should have an interface to support benchmarks that allocate a fresh environment for every run of the algorithm.

RyanGlScott:
Fixed in #136.

PR #136, opened 2017-04-25
Add new methods for allocating and cleaning up benchmark environments.

### LINKED ISSUE #79: `env` like function for freeing resources

merijn:
This patch adds support for per run benchmark allocation/cleanup of
environment, allocation/cleanup per batch of runs, and adds cleanup support for
the existing `env` function.

Fixes #79, #134, #135. Obsoletes #124.

### DIFF
--- a/Criterion.hs
+++ b/Criterion.hs
@@ -17,6 +17,11 @@ module Criterion
     -- * Creating a benchmark suite
     , Benchmark
     , env
+    , envWithCleanup
+    , perBatchEnv
+    , perBatchEnvWithCleanup
+    , perRunEnv
+    , perRunEnvWithCleanup
     , bench
     , bgroup
     -- ** Running a benchmark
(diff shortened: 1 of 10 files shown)


merijn:
I just realised I forgot to add tests for the cleanup in presence of exceptions, I'll add those ASAP.

merijn:
Added some tests, and in the process fixed #114, because that was messing up the results of the test output.

merijn:
Tests passed and random dependency things fixed!

RyanGlScott:
Thanks @merijn! I wanted to comment to let you know that I haven't forgotten about this. We're going to have a make a new `criterion` release soon anyway to address #140, and this is one of the things I want to merge in time for that release.

merijn:
Excellent!

PR #137, opened 2017-05-12
Use statistics-0.14

RyanGlScott:
`criterion` needs a bit of a makeover to work properly with the latest version of `statistics` (0.14).


RyanGlScott:
This is needed to unblock https://github.com/fpco/stackage/issues/2504 (i.e., ensure that `criterion` continues to be available on Stackage).

Issue #119, opened 2016-08-17
Stop depending on deprecated hastache package


RyanGlScott:
The [`hastache`](http://hackage.haskell.org/package/hastache) package is listed as deprecated on Hackage. We should look into replacing it with a more maintained equivalent.

The `hastache` Haddocks recommend the [`mustache`](http://hackage.haskell.org/package/mustache) package. I've never tried it, though.

RyanGlScott:
An issue: most of the actively maintained mustache template libraries on Hackage don't support GHC 7.4, which is the earliest version of GHC that we maintain support for in `criterion`:

* [`mustache-2.0.1`](http://hackage.haskell.org/package/mustache-2.0.1): GHC 7.8
* [`stache-0.1.6`](http://hackage.haskell.org/package/stache-0.1.6): GHC 7.8

rrnewton:
Now that 8.0 is out, if the policy is three versions of GHC that should remove GHC 7.4 from the window.  @bos, is that the policy?

bos:
You're welcome to nuke hastache support if it makes maintenance easier :-)

RyanGlScott:
Well, one issue is that I don't know if `mustache` or `stache` (or some other mustache templating library) can be used as drop-in replacement for `hastache`. But even if they currently can't, I'm sure we could devise a way to make it work (even if it takes some PRs to accomplish it).

On a less technical level, I'm somewhat loathe to simply drop support for GHC 7.4. `criterion` is the _de factor_ benchmarking library, after all, and one thing that's sometimes useful to do is to compare the performance of programs between GHC versions, and if you can't build `criterion` with a particular GHC version, you'll have to benchmark it out-of-process, which has its own problems.

Again, this issue could possibly be overcome with a PR that relaxed the minimum GHC requirements for some mustache library of choice. I don't know how many bleeding-edge GHC features `mustache` or `stache` are using.

rrnewton:
Actually, your argument applies doubly because from what you pasted earlier, @RyanGlScott, those alternative libs drop *7.6* support as well as 7.4.

Can we keep stumbling along with the deprecated hastache through the GHC 8.0 cycle and revisit this at 8.2?

RyanGlScott:
Sure, I don't see a need to fix this immediately. `hastache` still works fine for `criterion`'s needs, and to be honest, I'm not 100% convinced that any of the possible replacements above are worth migrating to yet.

RyanGlScott:
[...]

PR #139, opened 2017-05-18
Port to use statistics-0.14

phadej:
Works for GHC >= 7.8

On older GHC it won't as `Estimate` isn't `Data` anymore (cascading so that about everything isn't, yet is still `Generic`). There are three options

- Check if `Data` instances can be derived by hand in `statistics`
- Workaround `Data` requirement in [`mkGenericContext`](https://www.stackage.org/haddock/lts-8.14/hastache-0.6.1/Text-Hastache-Context.html)
- Drop support for GHC < 7.8 (that would be pity)
- Rewrite report generation using `blaze-markup` or `lucid` (but then we'll loose templating option. Is it used?)

comments?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 dist
+dist-newstyle
 cabal.sandbox.config
 .cabal-sandbox
 *~
--- a/Criterion/Analysis.hs
+++ b/Criterion/Analysis.hs
@@ -50,7 +50,8 @@ import Statistics.Regression (bootstrapRegress, olsRegress)
 import Statistics.Resampling (resample)
 import Statistics.Sample (mean)
 import Statistics.Sample.KernelDensity (kde)
-import Statistics.Types (Estimator(..), Sample)
+import Statistics.Resampling (Estimator(..))
+import Statistics.Types (Sample)
 import System.Random.MWC (GenIO)
 import qualified Data.List as List
 import qualified Data.Map as Map
@@ -58,7 +59,10 @@ import qualified Data.Vector as V
 import qualified Data.Vector.Generic as G
 import qualified Data.Vector.Unboxed as U
 import qualified Statistics.Resampling.Bootstrap as B
+import qualified Statistics.Types as B
 import Prelude
+
+type Estimate = B.Estimate B.ConfInt Double
 
 -- | Classify outliers in a data set, using the boxplot technique.
 classifyOutliers :: Sample -> Outliers
@@ -81,8 +85,8 @@ classifyOutliers sa = U.foldl' ((. outlier) . mappend) mempty ssa
 
 -- | Compute the extent to which outliers in the sample data affect
 -- the sample mean and standard deviation.
-outlierVariance :: B.Estimate  -- ^ Bootstrap estimate of sample mean.
-                -> B.Estimate  -- ^ Bootstrap estimate of sample
+outlierVariance :: Estimate  -- ^ Bootstrap estimate of sample mean.
+                -> Estimate  -- ^ Bootstrap estimate of sample
                                --   standard deviation.
                 -> Double      -- ^ Number of original iterations.
                 -> OutlierVariance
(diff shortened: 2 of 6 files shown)


Shimuuar:
Or rewrite `mkGenericContext` using generics?

phadej:
@Shimuuar that's one variant of "Workaround Data requirement in `mkGenericContext`" :)

RyanGlScott:
Thanks for doing this, @phadej.

What is the difference between #137 and this PR?

Also, I wasn't aware of the breakage on old GHCs. (the CI infrastructure doesn't test all supported versions of GHC... Hm.) Why does this `mkGenericContext` issue only affect old GHCs? Is there a way we can patch up `statistics` to fix this?

phadej:
Didn't notice your PR, it looks about the same (and better style!)

the problem is `Typeable` and `Data` for `Estimate`, which is now of kind `(* -> *) -> * -> *`:

```
Statistics/Types.hs:334:1:
    Can't make a derived instance of `Typeable (Estimate e a)':
      `Estimate' must only have arguments of kind `*'
```

RyanGlScott:
> the problem is `Typeable` and `Data` for `Estimate`, which is now of kind `(* -> *) -> * -> *`

Aw, fiddlesticks.

> I'm not sure if the instance for it (and `Data`) can be written by hand?

Believe it or not, there _is_ a way you could hand-write such a `Typeable` instance on old GHCs. For instance, I [do this in `base-orphans`](https://github.com/haskell-compat/base-orphans/blob/4637fb100ae6fadda4c9102bb764f811b30f867a/src/Data/Orphans.hs#L1175-L1179) to backport a `Typeable` instance for `WrappedMonad`, which has the same kind as `Estimate`:

```haskell
instance Typeable2 a => Typeable2 (WrappedArrow a) where
  typeOf2 t = mkTyConApp wrappedArrowTyCon [typeOf2 (a t)]
    where
      a :: WrappedArrow a b c -> a b c
      a = undefined
```

So the possibility is there.

A more radical option is to explore dropping support for GHCs earlier than 7.8. Aside from this particular issue with `hastache`, there's also the problem that `hastache` is deprecated and no longer maintained. I've been looking into [replacing `hastache` with a more modern templating library](https://github.com/bos/criterion/issues/119) like `mustache` or `stache`. The problem is that both libraries require GHC 7.8 as a minimum... but if we have to drop pre-7.8 support in `criterion` anyways, that would give us a better excuse to pursue replacing `hastache`.

BTW you asked earlier:

> (but then we'll loose templating option. Is it used?)

Yes. @erikd, for instance, uses this feature IIRC.

phadej:
I was able to wip-port `stache` from `megaparsec` to `parsec`. It compiles, yet have to port tests to see if it actually works :) https://github.com/phadej/microstache Would that be a solution?

phadej:
And `microstache` works: https://travis-ci.org/phadej/microstache/builds/233782212

RyanGlScott:
@phadej, wonderful! That gives us a `hastache` replacement and completely avoids the legacy GHC issue, which is wonderful.

The next step is to see if `microstache` can work as a suitable replacement for `hastache` in the ways that `criterion` uses it. I know that [`criterion` uses `hastache`'s support for lambdas](https://github.com/bos/criterion/blob/f117bec84c9bf460549fa565b8f30fac075a7db7/Criterion/Report.hs#L95), which is something that not all templating engine libraries support (e.g., `stache`), so that's something to watch out for.

phadej:
from `stache` (and `microstache`) readme:

> One feature that is not currently supported is lambdas. The feature is marked as optional in the spec and can be emulated via processing of parsed template representation. The decision to drop lambdas is intentional, for the sake of simplicity and better integration with Aeson.

RyanGlScott:
Oh right, you _did_ base this off of `stache`. Well, that's a bit awkward... do you know how difficult it is to do the "emulat[ion] via processing of parsed template representation" part?

phadej:
@RyanGlScott AFAICS, lambda is used to implement `include`. So we can just go thru the parsed `Node` AST, locate `include` `Node`s, and replace them with `includeFile` contents.

phadej:
Closed in favor of #141

PR #142, opened 2017-05-19
Add a program that creates reports from previously produced JSON files.

merijn:
Turn the non-functional criterion executable in a very useful, albeit it basic and rudimentary, tool to convert previously produced JSON files into those fancy-pants HTML reports.


merijn:
Solves #143

RyanGlScott:
Thanks, @merijn. The code generally looks fine. Some comments:

* Using the name `criterion` for this executable is perhaps a bit vague. I'm thinking we should rename it to something like `criterion-report` to better reflect its intended purpose.
* Can you update [the tutorial](https://github.com/bos/criterion/blob/3daefb899a925328c3d90268cf12a53eba0ae29d/www/tutorial.md) with an example of how you'd use this program?

criterion.cabal
```diff
@@ -111,27 +111,16 @@ library
...
 
--- [2016.05.30] RRN: I'm not sure where this was going.
--- It looks like it was meant for post-facto analysis of
--- reports?  Will have to ask BOS.
---------------------------------------------------------
--- executable criterion
---   hs-source-dirs: app
---   main-is:        App.hs
---   other-modules:
---     Options
```
RyanGlScott: The Travis build appears to be failing because you removed `other-modules: Options`.

merijn:
Yeah, in general the code needs to be cleaned up a little too, I'll do that sometime this weekend.

merijn:
Ok, fixed the missing other-modules and added a short bit to the tutorial. I have more interesting plans for the executable, but probably won't have time in the coming weeks to implement those.

Since you said you wanted to make a new release soon, it's probably best to keep the executable as-is, at the moment (i.e. limited to just generating reports of JSON output) and I'll start bugging people to merge something more interesting once I get around to finishing work on it.

phadej:
This broke the `master` on all GHC's for a bit different reasons: https://travis-ci.org/bos/criterion

merijn:
@phadej Yeah, I was relying on Travis for testing (due to lack of old GHCs locally), combined with a quick in-between meetings commit and a bit overeager merge that broke stuff. Should be fixed now that #144 is merged :)

PR #144, opened 2017-05-23
Fixes for the build problems introduced by the earlier gen-report merge.

merijn:
Quick commit in-between work and lack of old GHC's to test with lead to the previously merged gen-report breaking the Travis tests. This should fix all those issues. See #142


RyanGlScott:
Thanks!

PR #71, opened 2014-12-16
Add errorbars and enable navigation

Rufflewind:
- Add error bars showing the standard deviation (is there a better statistic to show?) using the errorbars plugin; fixes #6
- Allow panning and zooming using the navigate plugin (this can be useful if, say, parts of the graph are obscured by the legend)

PR #150, opened 2017-06-13
Add error bars (continuation of #71)

Rufflewind:
Continuation of https://github.com/bos/criterion/pull/71

### DIFF
--- a/Criterion/Report.hs
+++ b/Criterion/Report.hs
@@ -94,16 +94,18 @@ formatReport reports templateName = do
 
     jQuery <- T.readFile =<< JQuery.file
     flot <- T.readFile =<< Flot.file Flot.Flot
+    flotErrorbars <- T.readFile =<< Flot.file Flot.FlotErrorbars
 
     -- includes, only top level
     templates <- getTemplateDir
     template <- includeTemplate (includeFile [templates]) template0
 
     let context = object
-            [ "json"      .= reports
-            , "report"    .= map inner reports
-            , "js-jquery" .= jQuery
-            , "js-flot"   .= flot
+            [ "json"              .= reports
+            , "report"            .= map inner reports
+            , "js-jquery"         .= jQuery
+            , "js-flot"           .= flot
+            , "js-flot-errorbars" .= flotErrorbars
             ]
 
     return (renderMustache template context)
--- a/Criterion/Report.hs
+++ b/Criterion/Report.hs
@@ -95,6 +95,7 @@ formatReport reports templateName = do
     jQuery <- T.readFile =<< JQuery.file
     flot <- T.readFile =<< Flot.file Flot.Flot
     flotErrorbars <- T.readFile =<< Flot.file Flot.FlotErrorbars
+    flotNavigate <- T.readFile =<< Flot.file Flot.FlotNavigate
 
     -- includes, only top level
     templates <- getTemplateDir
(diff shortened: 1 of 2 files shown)


RyanGlScott:
This look great? A couple of questions:

* What is the best way to zoom in/out of the graph? Scrolling the mousewheel/trackpad seems to make it zoom, but that also causes the whole page to scroll up/down at the same time! Is there a cleaner way to accomplish this?
* Is there any way to make the zoom... less powerful? In particular, I managed to get it to a very warped state by accidentally zooming out a lot:

criterion

Rufflewind:
> Scrolling the mousewheel/trackpad seems to make it zoom, but that also causes the whole page to scroll up/down at the same time!

What browser are you testing it with?  Do you also run into that same problem with this page:
 http://www.flotcharts.org/flot/examples/navigate/index.html ?

On my Chromium, mousewheel just zooms in and out if your cursor is over the graph, and scrolls the window if the cursor is outside.

> Is there any way to make the zoom... less powerful?

This could be done by tweaking the `zoom.amount` attribute in the plot options ([default.tpl:318](https://github.com/bos/criterion/pull/150/commits/9ea9c2e6f872a2703ae63ed6828b7df808c4ff9d#diff-ca0bc1a3409ff865bbb3161972350affR318)). Default is `1.5`.  If you want something slower, could try `1.1` or `1.05`.

RyanGlScott:
> What browser are you testing it with? Do you also run into that same problem with this page:
http://www.flotcharts.org/flot/examples/navigate/index.html ?

Ah, that explains it. I experience the issue with Firefox, but not with Chromium.

> This could be done by tweaking the `zoom.amount` attribute in the plot options (default.tpl:318). Default is `1.5`. If you want something slower, could try `1.1` or `1.05`.

Ah, wonderful. `1.05` is much more to my liking (in my subjective opinion).

I'm still a bit unsettled that you're even _allowed_ to zoom out that much (and moreover, that you're allowed to move the camera to negative x- and y-axis values, which aren't used here), but perhaps that's just a quirk of `js-flot`...

Rufflewind:
Okay, I've updated the branch to use 1.05.

> I experience the issue with Firefox, but not with Chromium.

I tested it on Firefox (Linux) and it seems to work fine, so I wonder if it's some other reason (e.g. interaction with some Firefox extension/plugin)?  Maybe it should be reported as a bug in flot's repo?

RyanGlScott:
> I tested it on Firefox (Linux) and it seems to work fine, so I wonder if it's some other reason (e.g. interaction with some Firefox extension/plugin)? Maybe it should be reported as a bug in flot's repo?

That's pretty strange. I tested it on OS X, but I wouldn't rule out plugin weirdness either. I'll try Linux as well when I get home.

RyanGlScott:
Hrm, that's strange. I can confirm that scrolling works as expected on Firefox on Linux, but not on Windows (or OS X)!

Rufflewind:
I tested it on Firefox 54.0 (32-bit) in a Windows 10 VM and I cannot reproduce the problem.

RyanGlScott:
I've reported this at https://github.com/flot/flot/issues/1554.

RyanGlScott:
Hm. I can still reproduce the issue on Firefox 57 on macOS. But at the same time, I'd really hate to see this languish, since this is a nice addition. In light of this, I think I'll just go ahead and merge this and deal with the consequences. (If need be, we can always choose to revert this.)

Rufflewind:
Thanks!

PR #156, opened 2017-09-13
Add Semigroup instance for Outliers

bgamari:
This will be needed in GHC 8.4 due to the Semigroup/Monoid proposal.

### DIFF
--- a/Criterion/Types.hs
+++ b/Criterion/Types.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE Trustworthy #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE DeriveDataTypeable, DeriveGeneric, GADTs, RecordWildCards #-}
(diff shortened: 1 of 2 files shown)


criterion.cabal
```diff
@@ -92,6 +92,7 @@ library
     deepseq >= 1.1.0.0,
     directory,
     exceptions >= 0.8.2 && < 0.9,
+    semigroups,
```
RyanGlScott: Nitpick: Can you put this alphabetical order among the other dependencies?

Criterion/Types.hs
```diff
@@ -660,6 +661,9 @@ instance Binary OutlierEffect where
             _ -> fail $ "get for OutlierEffect: unexpected " ++ show i
 instance NFData OutlierEffect
 
+instance Semigroup Outliers where
+    (<>) = addOutliers
+
 instance Monoid Outliers where
     mempty  = Outliers 0 0 0 0 0
     mappend = addOutliers
```
RyanGlScott: While you're at it, can you change this to:

```haskell
#if !(MIN_VERSION_base(4,11,0))
     mappend = (<>)
#endif
```

This will help placate `-Wcompat`, and will make it forward compatible when we eventually drop `mappend` from `Monoid`.

bgamari:
Done.

PR #158, opened 2017-10-04
Added documentation for the `glob` match type.

### LINKED ISSUE #157: Document the usage of glob
I could not find any documentation of the use of glob patterns for selecting benchmarks. Luckily there was [a question](https://github.com/bos/criterion/issues/101) asked before that helped me. Is it possible to include this information somewhere?

Also including an example in the documentation of [`MatchType`](http://hackage.haskell.org/package/criterion-1.1.4.0/docs/Criterion-Main-Options.html#t:MatchType) might also help.

capitanbatata:
Fixes #157.

### DIFF
--- a/Criterion/Main/Options.hs
+++ b/Criterion/Main/Options.hs
@@ -54,7 +54,7 @@ data MatchType = Prefix
                  -- type benchmark names are treated as if they were
                  -- file-paths. For example, the glob patterns @"*/ba*"@ and
                  -- @"*/*"@ will match @"foo/bar"@, but @"*"@ or @"*bar"@
-                 -- __wil not__.
+                 -- __will not__.
                | Pattern
                  -- ^ Match by searching given substring in benchmark
                  -- paths.
--- a/Criterion/Main/Options.hs
+++ b/Criterion/Main/Options.hs
@@ -51,9 +51,9 @@ data MatchType = Prefix
                  -- @\"foo\"@ will match @\"foobar\"@.
                | Glob
                  -- ^ Match by Unix-style glob pattern. When using this match
-                 -- type benchmark names are treated as if they were
-                 -- file-paths. For example, the glob patterns @"*/ba*"@ and
-                 -- @"*/*"@ will match @"foo/bar"@, but @"*"@ or @"*bar"@
+                 -- type, benchmark names are treated as if they were
+                 -- file-paths. For example, the glob patterns @\"*/ba*\"@ and
+                 -- @\"*/*\"@ will match @\"foo/bar\"@, but @\"*\"@ or @\"*bar\"@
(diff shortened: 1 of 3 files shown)


www/tutorial.md
```diff
@@ -470,7 +470,10 @@ currently four ways to configure `--match`:
 * `--match prefix`: Check if the given string is a prefix of a benchmark
   path. For instance, `"foo"` will match `"foobar"`.
 
-* `--match glob`: Use the given string as a Unix-style glob pattern.
+* `--match glob`: Use the given string as a Unix-style glob pattern. Bear in
+  mind that performing a glob match on benchmarks names is done as if they were
+  file paths, so for instance both `"*/ba*"` and "*/*" will match "foo/bar",
```
RyanGlScott: Make sure to surround all of these strings in backticks (for consistency with the rest of the tutorial).

Criterion/Main/Options.hs
```diff
@@ -50,7 +50,11 @@ data MatchType = Prefix
                  -- ^ Match by prefix. For example, a prefix of
                  -- @\"foo\"@ will match @\"foobar\"@.
                | Glob
-                 -- ^ Match by Unix-style glob pattern.
+                 -- ^ Match by Unix-style glob pattern. When using this match
+                 -- type benchmark names are treated as if they were
+                 -- file-paths. For example, the glob patterns @"*/ba*"@ and
```
RyanGlScott: In Haddock, double quotes need to be escaped with a backslash (e.g., `@\"foo\"@`), or else they'll be rendered incorrectly.

Criterion/Main/Options.hs
```diff
@@ -50,7 +50,11 @@ data MatchType = Prefix
                  -- ^ Match by prefix. For example, a prefix of
                  -- @\"foo\"@ will match @\"foobar\"@.
                | Glob
-                 -- ^ Match by Unix-style glob pattern.
+                 -- ^ Match by Unix-style glob pattern. When using this match
+                 -- type benchmark names are treated as if they were
```
RyanGlScott: I would use a comma to separate "type" and "benchmark" here—it reads strangely without it.

www/tutorial.md
capitanbatata: Oops, I missed that. It's fixed now.

Criterion/Main/Options.hs
capitanbatata: Indeed. Added a comma.

Criterion/Main/Options.hs
capitanbatata: I missed that as well :/

RyanGlScott:
Thanks!

PR #175, opened 2018-01-06
Changed `measure` to properly compute statistics when using `perRunEnv`.

### LINKED ISSUE #174: `perRunEnv` yields too small result?
I wrote a benchmark using `perRunEnv` and got too small result.

Here's an example:

```haskell
module Main where
import Criterion.Main
import Control.Concurrent

main =
  do
    defaultMain
      [
        bench "whnfIO" $ whnfIO $ threadDelay 100000,
        bench "perRunEnv" $ perRunEnv (return ()) $ \_ -> threadDelay 100000
      ]
```

I run above code in linux environment and got:

```
benchmarking whnfIO
time                 100.3 ms   (100.2 ms .. 100.4 ms)
mean                 100.2 ms   (100.2 ms .. 100.3 ms)
std dev              78.44 μs   (63.88 μs .. 97.07 μs)

benchmarking perRunEnv
time                 19.46 ms   (14.38 ms .. 33.42 ms)
                     0.969 R²   (0.967 R² .. 1.000 R²)
mean                 36.17 ms   (25.07 ms .. 44.49 ms)
std dev              12.71 ms   (0.0 s .. 14.41 ms)
variance introduced by outliers: 73% (severely inflated)
```

I expect that `perRunEnv` should return a result near 100 ms like `whnfIO`.

merijn:
Fixes #174

### DIFF
--- a/Criterion/Measurement.hs
+++ b/Criterion/Measurement.hs
@@ -188,7 +188,7 @@ getGCStatistics = do
 measure :: Benchmarkable        -- ^ Operation to benchmark.
         -> Int64                -- ^ Number of iterations.
         -> IO (Measured, Double)
-measure bm iters = runBenchmarkable bm iters addResults $ \act -> do
+measure bm iters = runBenchmarkable bm iters addResults $ \ !n act -> do
   startStats <- getGCStatistics
   startTime <- getTime
   startCpuTime <- getCPUTime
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Awesome. Thanks, @merijn!

PR #176, opened 2018-01-09
Fix GC/RTS stats for base >= 4.10.0

avieth:
From `GHC.Stats` documentation for `GCStats` in base 4.10.0.0 on [hackage](https://hackage.haskell.org/package/base-4.10.0.0/docs/GHC-Stats.html):

> Statistics about memory usage and the garbage collector. Apart from currentBytesUsed and currentBytesSlop all are cumulative values since the program started.

For base >= 4.10.0 criterion chooses to use the new `RTSStats` type for GC measurements, but uses `GCDetails` for bytes allocated, bytes copied, and GC CPU times. This gave bogus results when, for instance, regressing allocated over iters. This program:

```Haskell
import Data.ByteString (ByteString)
import qualified Data.ByteString as BS
import Criterion
import Criterion.Main

makeBS :: Int -> ByteString
makeBS n = BS.replicate n 0

benchmarks :: [Benchmark]
benchmarks = [
    bench "make" (nf makeBS 1024)
  ]

main :: IO ()
main = defaultMain benchmarks
```

would show

```
allocated:           0.110 R²   (0.083 R² .. 0.145 R²)
  iters              0.414      (0.329 .. 0.539)
  y                  474831.157 (419995.999 .. 527923.352)
```

because the `GCDetails` gives "Details about the most recent GC".

With this patch the above program gives

```
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              1231.941   (1231.790 .. 1232.058)
  y                  -307784.572 (-341986.905 .. -274514.542)
```

### DIFF
--- a/Criterion/Measurement.hs
+++ b/Criterion/Measurement.hs
@@ -120,20 +120,20 @@ getGCStatistics = do
       nsToSecs ns = fromIntegral ns * 1.0E-9
 
   return $ Just GCStatistics {
-      gcStatsBytesAllocated         = fromIntegral $ gcdetails_allocated_bytes gcdetails
+      gcStatsBytesAllocated         = fromIntegral $ allocated_bytes stats
     , gcStatsNumGcs                 = fromIntegral $ gcs stats
     , gcStatsMaxBytesUsed           = fromIntegral $ max_live_bytes stats
     , gcStatsNumByteUsageSamples    = fromIntegral $ major_gcs stats
     , gcStatsCumulativeBytesUsed    = fromIntegral $ cumulative_live_bytes stats
-    , gcStatsBytesCopied            = fromIntegral $ gcdetails_copied_bytes gcdetails
+    , gcStatsBytesCopied            = fromIntegral $ copied_bytes stats
     , gcStatsCurrentBytesUsed       = fromIntegral $ gcdetails_live_bytes gcdetails
     , gcStatsCurrentBytesSlop       = fromIntegral $ gcdetails_slop_bytes gcdetails
     , gcStatsMaxBytesSlop           = fromIntegral $ max_slop_bytes stats
     , gcStatsPeakMegabytesAllocated = fromIntegral (max_mem_in_use_bytes stats) `quot` (1024*1024)
     , gcStatsMutatorCpuSeconds      = nsToSecs $ mutator_cpu_ns stats
     , gcStatsMutatorWallSeconds     = nsToSecs $ mutator_elapsed_ns stats
-    , gcStatsGcCpuSeconds           = nsToSecs $ gcdetails_cpu_ns gcdetails
-    , gcStatsGcWallSeconds          = nsToSecs $ gcdetails_elapsed_ns gcdetails
+    , gcStatsGcCpuSeconds           = nsToSecs $ cpu_ns stats
+    , gcStatsGcWallSeconds          = nsToSecs $ elapsed_ns stats
     , gcStatsCpuSeconds             = nsToSecs $ cpu_ns stats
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Oh dear. Thanks for noticing this!

I'll admit that I transitioned from the old `GHC.Stats` interface to the new one by following a similar commit in `accelerate` (see this commit: https://github.com/AccelerateHS/accelerate/commit/650ef2d9247cf8889b13edb148520a65b6931fa7#diff-e7b3ae196db40f076ad137cdcdad0027). So if the code in `criterion` is wrong, there's a good chance that `accelerate` is wrong, too!

cc @tmcdonell

tmcdonell:
Thanks for the heads up! I will check it out.

RyanGlScott:
Indeed, looking closer at the source code for `GHC.Stats` makes it evident that some of what `criterion` is doing is incorrect:

* Compare [`bytesAllocated`](https://github.com/ghc/ghc/blob/0156a3d815b784510a980621fdcb9c5b23826f1e/libraries/base/GHC/Stats.hsc#L285) to [`allocated_bytes`](https://github.com/ghc/ghc/blob/0156a3d815b784510a980621fdcb9c5b23826f1e/libraries/base/GHC/Stats.hsc#L162) (not [`gcdetails_allocated_bytes`](https://github.com/ghc/ghc/blob/0156a3d815b784510a980621fdcb9c5b23826f1e/libraries/base/GHC/Stats.hsc#L183), which is from a different struct).
* Compare [`bytesCopied`](https://github.com/ghc/ghc/blob/0156a3d815b784510a980621fdcb9c5b23826f1e/libraries/base/GHC/Stats.hsc#L290) to [`copied_bytes`](https://github.com/ghc/ghc/blob/0156a3d815b784510a980621fdcb9c5b23826f1e/libraries/base/GHC/Stats.hsc#L169) (not [`gcdetails_copied_bytes`](https://github.com/ghc/ghc/blob/0156a3d815b784510a980621fdcb9c5b23826f1e/libraries/base/GHC/Stats.hsc#L190), which is from a different struct).

However, it appears this patch gets two things wrong.

* Compare [`gcCpuSeconds`](https://github.com/ghc/ghc/blob/0156a3d815b784510a980621fdcb9c5b23826f1e/libraries/base/GHC/Stats.hsc#L303) to [`gc_cpu_ns`](https://github.com/ghc/ghc/blob/0156a3d815b784510a980621fdcb9c5b23826f1e/libraries/base/GHC/Stats.hsc#L175). This patch, however, is trying to use `cpu_ns`.
* Compare [`gcWallSeconds`](https://github.com/ghc/ghc/blob/0156a3d815b784510a980621fdcb9c5b23826f1e/libraries/base/GHC/Stats.hsc#L304) to [`gc_elapsed_ns`](https://github.com/ghc/ghc/blob/0156a3d815b784510a980621fdcb9c5b23826f1e/libraries/base/GHC/Stats.hsc#L176). This patch, however, is trying to use `elapsed_ns`.

@avieth, do you agree with this assessment?

avieth:
> do you agree with this assessment?

Oops yes you're right, it should be `gc_cpu_ns` and `gc_elapsed_ns`. I'll fix that.

What do you think about the movement of `getGCStatistics`? I thought perhaps the intermediate

```Haskell
    endTime <- getTime
    endCpuTime <- getCPUTime
    endCycles <- getCycles
```

would taint the results but maybe I'm wrong about this.

RyanGlScott:
I think the order you're using now makes much more sense, since it performs the actions in the order they were originally performed.

PR #177, opened 2018-01-09
performGC before getting RTS/GC stats

avieth:
performGC before getting RTS/GC stats

The deprecated getGCStats had some nice documentation: "If you would
like your statistics as recent as possible, first run a performGC"

This is sadly missing from getRTSStats but I believe it still holds.
When regressing allocated over iters, I'd see a believable slope, but an
unbelievable y intercept, something like -300000 when the slope is 150
or so. That's because the first hundred or so measurements were 0 for
bytes allocated, as the RTS didn't bother to run a GC before the RTS
stats were sampled.

Now we do 3 samples:

  1. performGC before the first one, to ensure it's up-to-date.
  2. Do the second one after the action, without a performGC, so we can
     get legit readings on the GC-related stats.
  3. performGC and then sample again, so we can get up-to-date readings
     on other metrics.
  4. Carefully choose whether to diff start stats against the end stats
     per- or post-GC.

Also included is a fix to the ToJSON Measurements instance, which
duplicated the mutator cpu seconds where GC cpu seconds should go.

### DIFF
--- a/Criterion/Measurement.hs
+++ b/Criterion/Measurement.hs
@@ -337,7 +337,7 @@ applyGCStatistics (Just endPostGC) (Just endPreGC) (Just start) m = m {
   , measMutatorCpuSeconds  = diff endPostGC start gcStatsMutatorCpuSeconds
   , measGcWallSeconds      = diff endPreGC  start gcStatsGcWallSeconds
   , measGcCpuSeconds       = diff endPreGC  start gcStatsGcCpuSeconds
-  } where diff end start f = f end - f start
+  } where diff a b f = f a - f b
 applyGCStatistics _ _ _ m = m
 
 -- | Convert a number of seconds to a string.  The string will consist
--- a/Criterion/Measurement.hs
+++ b/Criterion/Measurement.hs
@@ -108,7 +108,7 @@ data GCStatistics = GCStatistics
 -- | Try to get GC statistics, bearing in mind that the GHC runtime
 -- will throw an exception if statistics collection was not enabled
 -- using \"@+RTS -T@\".
--- If you need guaranteed up-to-date stats, call performGC first.
+-- If you need guaranteed up-to-date stats, call 'performGC' first.
 getGCStatistics :: IO (Maybe GCStatistics)
 #if MIN_VERSION_base(4,10,0)
 -- Use RTSStats/GCDetails to gather GC stats
--- a/Criterion/Measurement.hs
+++ b/Criterion/Measurement.hs
@@ -108,6 +108,7 @@ data GCStatistics = GCStatistics
 -- | Try to get GC statistics, bearing in mind that the GHC runtime
 -- will throw an exception if statistics collection was not enabled
 -- using \"@+RTS -T@\".
+-- If you need guaranteed up-to-date stats, call performGC first.
 getGCStatistics :: IO (Maybe GCStatistics)
 #if MIN_VERSION_base(4,10,0)
 -- Use RTSStats/GCDetails to gather GC stats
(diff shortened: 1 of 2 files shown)


Criterion/Measurement.hs
```diff
@@ -108,6 +108,7 @@ data GCStatistics = GCStatistics
 -- | Try to get GC statistics, bearing in mind that the GHC runtime
 -- will throw an exception if statistics collection was not enabled
 -- using \"@+RTS -T@\".
+-- If you need guaranteed up-to-date stats, call performGC first.
```
RyanGlScott: Put singlequotes around `performGC` (i.e., `'performGC'`) so that Haddock will link it.

Criterion/Measurement.hs
```diff
@@ -313,22 +321,24 @@ measured = Measured {
 -- | Apply the difference between two sets of GC statistics to a
 -- measurement.
 applyGCStatistics :: Maybe GCStatistics
-                  -- ^ Statistics gathered at the __end__ of a run.
+                  -- ^ Statistics gathered at the __end__ of a run, post GC.
+                  -> Maybe GCStatistics
+                  -- ^ Statistics gathered at the __end__ of a run, pre GC.
```
RyanGlScott: The order here (first post-GC statistics, then pre-GC statistics) feels all wonky to me. Why not have the first argument be pre-GC statistics, then following by the post-GC statistics? We're going to need a breaking change anyway, so we might as well set things straight in the process.

Criterion/Measurement.hs
```diff
@@ -313,22 +321,24 @@ measured = Measured {
...
-applyGCStatistics _ _ m = m
+applyGCStatistics (Just endPostGC) (Just endPreGC) (Just start) m = m {
+    measAllocated          = diff endPostGC start gcStatsBytesAllocated
+  , measNumGcs             = diff endPreGC  start gcStatsNumGcs
+  , measBytesCopied        = diff endPostGC start gcStatsBytesCopied
+  , measMutatorWallSeconds = diff endPostGC start gcStatsMutatorWallSeconds
+  , measMutatorCpuSeconds  = diff endPostGC start gcStatsMutatorCpuSeconds
+  , measGcWallSeconds      = diff endPreGC  start gcStatsGcWallSeconds
+  , measGcCpuSeconds       = diff endPreGC  start gcStatsGcCpuSeconds
+  } where diff a b f = f a - f b
```
RyanGlScott: This `b` argument will always be `start`, yes? Why not keep it inline in `diff`?

Criterion/Measurement.hs
```diff
@@ -313,22 +321,24 @@ measured = Measured {
...
-  } where diff f = f end - f start
-applyGCStatistics _ _ m = m
+applyGCStatistics (Just endPostGC) (Just endPreGC) (Just start) m = m {
+    measAllocated          = diff endPostGC start gcStatsBytesAllocated
+  , measNumGcs             = diff endPreGC  start gcStatsNumGcs
+  , measBytesCopied        = diff endPostGC start gcStatsBytesCopied
+  , measMutatorWallSeconds = diff endPostGC start gcStatsMutatorWallSeconds
+  , measMutatorCpuSeconds  = diff endPostGC start gcStatsMutatorCpuSeconds
+  , measGcWallSeconds      = diff endPreGC  start gcStatsGcWallSeconds
+  , measGcCpuSeconds       = diff endPreGC  start gcStatsGcCpuSeconds
```
RyanGlScott: In the comments, can you include a justification for which things are `diff`ed with `endPreGC`, and which things are `diff`ed with `endPostGC`?

Criterion/Measurement.hs
avieth: My reasoning was that the "later" argument always came first, so I kept with the trend. It doesn't matter so much to me so I'll change it up if you insist.

Criterion/Measurement.hs
avieth: Will do

avieth:
> Thanks for noticing this, @avieth. It's unfortunate timing that this requires another breaking change after we just released criterion-1.3.0.0, but I suppose that's just how it is sometimes.

Then again, apparently I'm the only person who uses regressions on the GC stats :) else someone would probably have noticed this by now.

Criterion/Measurement.hs
RyanGlScott: Ah, that's a good point I hadn't considered. In that case, your order is fine, since we'd want to preserve the existing convention.

RyanGlScott:
Thanks!

PR #182, opened 2018-02-01
Fix per-run combine time (fixes #181)

### LINKED ISSUE #181: Per-run time combining function is incorrect
When benchmarking with `perRun = True`, the `measure` function is incorrectly combining time taken.

https://github.com/bos/criterion/blob/52ef4a7699237a8952475146c51fc5d8ebff5876/Criterion/Measurement.hs#L197-L200

The second value returned by `measure` should be the absolute time, not an accumulation of time taken. You can see how it is used in `runBenchmark`:

https://github.com/bos/criterion/blob/52ef4a7699237a8952475146c51fc5d8ebff5876/Criterion/Measurement.hs#L270-L291

`loop` records a `start` time, and then each measurement reports when it ended. Once `loop` has run for longer than `timeLimit`, it can return. Adding `endTime`s together artificially inflates the time `loop` thinks it has taken, causing it to end early.

To fix this, we simply have to change `d1 + d2` to `d2`.

patrickdoc:
Before change:

    benchmarking perRun
    measurement took 231495 s (60 hours!!!)
    analysing with 1000 resamples
    bootstrapping with 13 of 22 samples (59%)
    time                 3.112 ms   (3.102 ms .. 3.122 ms)
    mean                 3.119 ms   (3.111 ms .. 3.122 ms)
    std dev              10.00 μs   (5.617 μs .. 16.74 μs)
    found 1 outliers among 13 samples (7.7%)
      1 (7.7%) low severe
    variance introduced by outliers: 4% (slightly inflated)

After:

    benchmarking perRun
    measurement took 5.135 s
    analysing with 1000 resamples
    bootstrapping with 33 of 42 samples (78%)
    time                 3.120 ms   (3.116 ms .. 3.124 ms)
    mean                 3.117 ms   (3.111 ms .. 3.120 ms)
    std dev              11.95 μs   (6.974 μs .. 20.56 μs)
    found 3 outliers among 33 samples (9.1%)
      3 (9.1%) low severe
    variance introduced by outliers: 2% (slightly inflated)

The reported time is now much more accurate, and we get double the number samples.

### DIFF
--- a/Criterion/Measurement.hs
+++ b/Criterion/Measurement.hs
@@ -197,7 +197,7 @@ measure bm iters = runBenchmarkable bm iters addResults $ \ !n act -> do
   return (m, endTime)
   where
     addResults :: (Measured, Double) -> (Measured, Double) -> (Measured, Double)
-    addResults (!m1, !d1) (!m2, !d2) = (m3, d1 + d2)
+    addResults (!m1, _) (!m2, !d2) = (m3, d2)
       where
         add f = f m1 + f m2
 
--- a/Criterion/Measurement.hs
+++ b/Criterion/Measurement.hs
@@ -196,6 +196,8 @@ measure bm iters = runBenchmarkable bm iters addResults $ \ !n act -> do
            }
   return (m, endTime)
   where
+    -- When combining runs, the Measured value is accumulated over many runs,
+    -- but the Double value is the most recent absolute measurement of time.
     addResults :: (Measured, Double) -> (Measured, Double) -> (Measured, Double)
     addResults (!m1, _) (!m2, !d2) = (m3, d2)
       where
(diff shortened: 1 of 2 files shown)


Criterion/Measurement.hs
```diff
@@ -197,7 +197,7 @@ measure bm iters = runBenchmarkable bm iters addResults $ \ !n act -> do
   return (m, endTime)
   where
     addResults :: (Measured, Double) -> (Measured, Double) -> (Measured, Double)
-    addResults (!m1, !d1) (!m2, !d2) = (m3, d1 + d2)
+    addResults (!m1, _) (!m2, !d2) = (m3, d2)
```
RyanGlScott: It might also be wise to add a comment describing why we add the `Measured`s together, but don't add the times together.

patrickdoc:
Done and done. I promise I will remember the changelog in the future 😣

Issue #189, opened 2018-03-17
Split out criterion-core ?


ocramz:
I would like a minimalistic performance library, that doesn't depend on HTML or javascript libraries (namely, `microstache`, `js-*`), or require processing of command line arguments (i.e. `optparse-applicative`). The main `criterion` library would then import `criterion-core` and re-add the bells and whistles.

Additionally, other profiling libraries could just import `criterion-core`, without incurring a large transitive dependency cost. The burden of run-time analysis and presentation would be transferred to these consumer libraries.

I'm willing to re-arrange the codebase in this fashion, would you accept such a PR? 
Thank you

RyanGlScott:
While I can appreciate the desire to minimize dependencies, I'm not sure how far you're going to get by just removing the HTML and CLI argument-parsing bits. Here's the dependency graph for `criterion`:

![deps](https://user-images.githubusercontent.com/2364661/37556165-5d469446-29c8-11e8-853c-a987d048f522.png)

The bulk of `criterion`'s dependencies come from `statistics`, which is actually used in the "core" of `criterion`.

Moreover, I'm not sure how feasible this idea even is. Most consumers use `criterion` by way of `defaultMain`, which is intended to be used for producing a full executable. By necessity, that must include the "bells and whistles". Or am I missing something?

RyanGlScott:
> @RyanGlScott Thank you for the clarifications and the graph, which is very enlightening. (BTW what did you use to produce it?)

I used [`cabal-plan`](http://hackage.haskell.org/package/cabal-plan):

```
$ cabal get criterion-1.4.0.0                                         
$ cd criterion-1.4.0.0/                                                                                                    
$ echo "packages: ." > cabal.project
$ cabal new-build --disable-tests --disable-benchmarks
$ cabal-plan --builddir=dist-newstyle --hide-builtin dot --tred --tred-weights | dot -Tpng -odeps.png
```

> In this picture, `criterion-core` could get rid of `statistics` as well, and only provide combinators for raw timing measurements on single functions (i.e. `Criterion.Measurement` and its direct dependencies).

`Criterion.Measurement` has `statistics` as a direct dependency, actually, since it imports `Criterion.Types` (which in turns uses `statistics`).

> I think It's not inconceivable that some consumer libraries might want to plug in alternative statistical functionality, or use the raw data in novel ways.

[...]

PR #191, opened 2018-03-27
Move measurement functionality to criterion-measurement

ocramz:
A separate measurement package with minimal dependencies (as discussed in #189 ).
Nothing changes in the interface, just some functionality is moved out into `criterion-measurement` and reimported in `criterion`
All tests pass.

Changes:

* Move out `cbits/` , `Criterion.Measurement`, `Criterion.Types.Internal`
* Move out parts of `Criterion.Types`
* Version bump criterion 1.4 -> 1.5


ocramz:
Travis complains because 'criterion-measurement' is not yet on Hackage; I actually don't know how to convince Cabal to build against a local package. So far I just used the 'packages' stanza in stack.yaml

Criterion/Types.hs
```diff
@@ -127,490 +128,14 @@ data Config = Config {
...
-
-measure :: (U.Unbox a) => (Measured -> a) -> V.Vector Measured -> U.Vector a
-measure f v = U.convert . V.map f $ v
+
+
+
+
+
+
+
```
RyanGlScott: There's a _lot_ of unnecessary line breaks here. Let's trim them down.

criterion-measurement/stack.yaml
```diff
@@ -0,0 +1,8 @@
+resolver: lts-10.5
```
RyanGlScott: Why do you need a separate `stack.yaml` file just for `criterion-measurement`? Doesn't the top-level one include this information already?

criterion-measurement/criterion-measurement.cabal
```diff
@@ -0,0 +1,68 @@
...
+    if os(windows)
+      c-sources: cbits/time-windows.c
+    else
+      c-sources: cbits/time-posix.c
+  }
+
+
+source-repository head
+  type:     git
+  location: https://github.com/bos/criterion
```
RyanGlScott: Put:

```
  subdir: criterion-measurement
```

as another stanza here.

criterion-measurement/criterion-measurement.cabal
```diff
@@ -0,0 +1,68 @@
...
+  default: False
+  manual: True
+
+library
+  hs-source-dirs:      src
+  exposed-modules:     Criterion.Measurement
+                       Criterion.Types
+  other-modules:       Criterion.Types.Internal
+  build-depends:       aeson >= 0.8
+                     , base >= 4.7 && < 5
```
RyanGlScott: Why are you restricting the lower version bounds of `base` here? (Note that in `criterion` itself, the lower bound is 4.5.)

criterion-measurement/criterion-measurement.cabal
```diff
@@ -0,0 +1,68 @@
...
+  ghc-options: -Wall -funbox-strict-fields
+  if impl(ghc >= 6.8)
+    ghc-options: -fwarn-tabs
+  if flag(fast)
+    ghc-options: -O0
+  else
+    ghc-options: -O2
+
+
+  c-sources: cbits/cycles.c
```
RyanGlScott: Since the `cbits` stuff lives in `criterion-measurement` now, we should be able to remove the corresponding `cbits` parts from `criterion.cabal`, yes? (Also, we should be able to remove the top-level `cbits` directory, since they should now live solely in `criterion-measurement/cbits`, IIUC.)

Criterion/Measurement.hs
```diff
@@ -1,9 +1,4 @@
-{-# LANGUAGE DeriveDataTypeable #-}
-{-# LANGUAGE DeriveGeneric #-}
-{-# LANGUAGE Trustworthy #-}
-{-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE BangPatterns, CPP, ForeignFunctionInterface,
-    ScopedTypeVariables #-}
+{-# LANGUAGE PackageImports #-}
```
RyanGlScott: Currently, there are two `Criterion.Measurement` files—one in `criterion`, and one in `criterion-measurement`—but they both export the exact same API. Moreover, the existence of both requires the use of `PackageImports` invasively, which I'm not fond of. I'd opt to just get rid of the `Criterion.Measurement` from `criterion` entirely. (It's a breaking change, after all.)

Criterion/Types.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE Trustworthy #-}
```
RyanGlScott: Similarly, we now have two different `Criterion.Types` modules across the two packages. But this time, the one in `criterion` exposes _more_ things than the one in `criterion-measurement`!

I'd favor renaming the version in `criterion-measurement` to `Criterion.Measurement.Types` to avoid this confusion. Similarly, rename `Criterion.Types.Internal` in `criterion-measurement` to `Criterion.Measurement.Types.Internal`.

criterion-measurement/src/Criterion/Types/Internal.hs
```diff
@@ -0,0 +1,60 @@
+{-# LANGUAGE BangPatterns #-}
```
RyanGlScott: Now that this lives here, there's no need for the `Criterion.Types.Internal` in `criterion`, is there? Delete it.

criterion-measurement/criterion-measurement.cabal
```diff
@@ -0,0 +1,68 @@
+name:                criterion-measurement
+version:             0.1.0.0
+synopsis:            Criterion measurement functionality and associated types
+description:         Measurement-related functionality extracted from Criterion, with minimal dependencies. The rationale for this is to enable alternative analysis front-ends.
+homepage:            https://github.com/bos/criterion
+license:             BSD3
+license-file:        LICENSE
```
RyanGlScott: There's not a `LICENSE` file alongside `criterion-measurement.cabal`, which is causing build failures on Travis.

criterion-measurement/stack.yaml
ocramz: I understand it's replicated information, but I'd rather keep this stack.yaml in case other downstream libraries need to build on it.

RyanGlScott: I don't understand. If anyone wants to clone `criterion` locally and use `stack` to build it, then they'll use the `stack.yaml` from the top-level directory, yes? Why is this ever needed?

criterion-measurement/criterion-measurement.cabal
ocramz: Done

criterion-measurement/criterion-measurement.cabal
ocramz: Added copy of main license

criterion-measurement/stack.yaml
ocramz: I thought this might be needed in order to build `criterion-measurement` as a standalone library? Or, I guess I didn't consider the possibility you're mentioning.

RyanGlScott: I'm not sure we're on the same page here, so let me try and contextualize this.

First, we never ship `stack.yaml`/`cabal.project` files in sdists that are uploaded to Hackage. The only reason one would ever use them is if one were to check out `criterion` locally on build it from source.

With that in mind, one should only ever need the `stack.yaml` file in the top-level directory, as it has enough information to build `criterion`, `criterion-measurement`, or both (depending on what targets you specify).

In light of this, the `stack.yaml` file in `criterion-measurement/` is entirely redundant. All of the information that `stack` needs to build it is already contained in the top-level `stack.yaml` file. Does this make sense?

ocramz: Absolutely, thank you; I'll remove the additional stack.yaml.

ocramz:
@RyanGlScott there is still an error with the cabal.project file generated in Travis:

```
The package location glob 'criterion-measurement-*/*.cabal' does not match any files or directories.
```

RyanGlScott:
This is because the sdist wasn't created due to an earlier build failure. Of particular note are [these lines](https://travis-ci.org/bos/criterion/jobs/358915476#L662):

```
cabal: Error: Could not find module: Criterion.Measurement.Types with any
suffix: ["gc","chs","hsc","x","y","ly","cpphs","hs","lhs","hsig","lhsig"]. If
the module is autogenerated it should be added to 'autogen-modules'.
```

You expose `Criterion.Measurement.Types` in `criterion-measurement.cabal`, but there isn't a file with that name anywhere.

criterion-measurement/src/Criterion/Measurement.hs
```diff
@@ -397,3 +398,5 @@ foreign import ccall unsafe "criterion_gettime" getTime :: IO Double
 -- | Return the amount of elapsed CPU time, combining user and kernel
 -- (system) time into a single measure.
 foreign import ccall unsafe "criterion_getcputime" getCPUTime :: IO Double
+
+
```
RyanGlScott: Please remove these unnecessary line breaks.

Criterion.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE PackageImports #-}
```
RyanGlScott: There's no need for `PackageImports` here.

tests/Tests.hs
```diff
@@ -1,6 +1,7 @@
+{-# LANGUAGE PackageImports #-}
```
RyanGlScott: There's no need for `PackageImports` here.

criterion.cabal
```diff
@@ -216,7 +205,8 @@ test-suite cleanup
     directory,
     tasty,
     tasty-hunit
-
+
```
RyanGlScott: Please remove this trailing whitespace.

Criterion/Main/Options.hs
```diff
@@ -27,8 +27,7 @@ import Data.Semigroup ((<>))
 
 import Control.Monad (when)
 import Criterion.Analysis (validateAccessors)
-import Criterion.Types (Config(..), Verbosity(..), measureAccessors,
-                        measureKeys)
+import Criterion.Types (Config(..), Verbosity(..), measureAccessors, measureKeys)
```
RyanGlScott: In general, I'd prefer not to clutter up the diff with minor aesthetic changes like this unless there's a good reason for them (which there doesn't seem to be here).

Criterion/Monad.hs
```diff
@@ -19,7 +19,7 @@ module Criterion.Monad
 import Control.Monad.Reader (asks, runReaderT)
 import Control.Monad.Trans (liftIO)
 import Criterion.Monad.Internal (Criterion(..), Crit(..))
-import Criterion.Types hiding (measure)
+import Criterion.Types
```
RyanGlScott: In general, I'd prefer not to clutter up the diff with minor aesthetic changes like this unless there's a good reason for them (which there doesn't seem to be here).

Criterion/Types.hs
```diff
@@ -60,7 +57,10 @@ module Criterion.Types
     , nf
     , whnf
     , nfIO
-    , whnfIO
+    , whnfIO          
+    -- * Configuration
```
RyanGlScott: Is there a reason the configuration stuff was moved in the export list?

criterion-measurement/src/Criterion/Measurement.hs
```diff
@@ -32,9 +32,10 @@ module Criterion.Measurement
     , measured
     , applyGCStatistics
     , threshold
-    ) where
+    )
+    where
```
RyanGlScott: In general, I'd prefer not to clutter up the diff with minor aesthetic changes like this unless there's a good reason for them (which there doesn't seem to be here).

ocramz:
All green now, thanks for your help @RyanGlScott

Criterion.hs
```diff
@@ -41,9 +41,9 @@ import Control.Monad (void)
 import Criterion.IO.Printf (note)
 import Criterion.Internal (runAndAnalyseOne)
 import Criterion.Main.Options (defaultConfig)
-import Criterion.Measurement (initializeTime)
 import Criterion.Monad (withConfig)
 import Criterion.Types
+import Criterion.Measurement (initializeTime)
```
RyanGlScott: Move this back to make the diff less noisy.

Criterion/Main/Options.hs
```diff
@@ -28,7 +28,7 @@ import Data.Semigroup ((<>))
 import Control.Monad (when)
 import Criterion.Analysis (validateAccessors)
 import Criterion.Types (Config(..), Verbosity(..), measureAccessors,
-                        measureKeys)
+                        measureKeys)
```
RyanGlScott: Trailing whitespace.

criterion.cabal
```diff
@@ -216,7 +205,8 @@ test-suite cleanup
...
```
RyanGlScott: Trailing whitespace.

tests/Tests.hs
```diff
@@ -1,6 +1,6 @@
 module Main (main) where
 
-import Criterion.Types
+import Criterion.Types
```
RyanGlScott: Trailing whitespace.

Criterion.hs
ocramz: What do you mean? Where should I move this?

tests/Tests.hs
ocramz: Emacs doesn't show any extra whitespace there.

Criterion.hs
RyanGlScott: This PR needlessly moves this import two lines down. Just put it back to where it originally was (as shown in the diff).

tests/Tests.hs
RyanGlScott: You fixed this in https://github.com/bos/criterion/pull/191/commits/9f5fe87b8746cce56b45b18508911598c03f1e57#diff-0bf78663781e2a6812584ddf0032c386R3, so that part is good now.

RyanGlScott:
LGTM. Thanks!

Issue #195, opened 2018-06-12
runBenchmark hangs on macOS


doomrobo:
There's some weird platform-specific behavior that I ran into recently. Using the standard `defaultMain` works just fine on macOS, but if I call `runBenchmark` (exposed by `Criterion.Measurement`) directly, my benchmarks don't terminate after >30min of running. Even weirder, when I call `defaultMain` *followed by* a `runBenchmark`, they both terminate.

I wasn't able to replicate this on my Linux machine, but I was able to replicate it on 2 macOS machines (both High Sierra). I also observed it for criterion versions 1.4.0 and 1.1.4.

I have a minimal testcase in a repo [here](https://github.com/doomrobo/criterion-hang-test). This is all the code necessary. It just runs a benchmark on the `id` function.

```haskell
module Main where

import qualified Criterion as C
import qualified Criterion.Main as C
import qualified Criterion.Measurement as C

main :: IO ()
main = do
  let b =  C.nf id (0 :: Int)
  -- If you uncomment the line below, the program will terminate within 30s
  --C.defaultMain [C.bench "id" b]
  x <- C.runBenchmark b 0.0
  print x
```

Any corrections or workarounds would be much appreciated.

RyanGlScott:
Quite strange. I don't have access to macOS, so there's not much I can do on my end to debug this, unfortunately. All I can offer is advice on where to look:

* Does the optimization setting affect this? i.e., does it make a difference whether the program is compiled with `-O0` or `-O`?
* Have you tried putting trace statements inside of `criterion` to narrow down which line of code is hanging?

doomrobo:
I figured it out. The loop inside `runBenchmark` calls `getTime` repeatedly to gauge how long the bench is taking. But `runBenchmark` never calls `initializeTime`! I think this is where the platform-specific behavior came from. Compare the C code for timing functions:

Linux version:
```C
void criterion_inittime(void)
{
}

double criterion_gettime(void)
{
    struct timespec ts;

    clock_gettime(CLOCK_MONOTONIC, &ts);

    return ts.tv_sec + ts.tv_nsec * 1e-9;
}
```

macOS version:
```C
void criterion_inittime(void)
{
    if (timebase_recip == 0) {
	mach_timebase_info(&timebase_info);
	timebase_recip = (timebase_info.denom / timebase_info.numer) / 1e9;
    }
}

double criterion_gettime(void)
{
    return mach_absolute_time() * timebase_recip;
}
```

[...]

PR #196, opened 2018-06-16
Call initializeTime before getTime in Measurement.hs; fixes issue #195

doomrobo:
My comment before was slightly mistaken. As you said, it's already explicitly stated in the docs that `initializeTime` must be called before using `getTime`.

So the changes I made were just adding `initializeTime` in the appropriate places. Let me know if there's anything I missed.

### DIFF
--- a/examples/Overhead.hs
+++ b/examples/Overhead.hs
@@ -11,6 +11,7 @@ import GHC.Stats as GHC
 
 main :: IO ()
 main = do
+  M.initializeTime -- Need to do this before calling M.getTime
   statsEnabled <- getRTSStatsEnabled
   defaultMain $ [
       bench "measure" $            whnfIO (M.measure (whnfIO $ return ()) 1)
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,9 @@
   This will let other libraries, e.g. alternative statistical analysis front-ends
   to import the measurement functionality alone as a lightweight dependency.
 
+* Fixed a macOS-specific bug where using `runAndAnalyse` and other lower-level
+  benchmarking functions would result in an infinite loop.
+
 1.4.1.0
 * Use `base-compat-batteries`.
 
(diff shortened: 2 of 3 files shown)


examples/Overhead.hs
```diff
@@ -11,6 +11,7 @@ import GHC.Stats as GHC
 
 main :: IO ()
 main = do
+  M.initializeTime -- Need to do this before calling M.getTime
```
RyanGlScott: I'm confused as to why this explicit call to `initializeTime` is necessary. Doesn't `defaultMain` call it already?

doomrobo: Yes, but it's not documented that `defaultMain` calls `initializeTime`. I think this example shows correct precautionary use.

RyanGlScott:
> Sure, I'll add a changelog comment. What is the next version number?

Just add it under 1.5.0.0, which should already be present in the changelog.

examples/Overhead.hs
RyanGlScott: Still, it feels strange that we'd have a call to `initializeTime` for no benefit. Perhaps we should at least document in `initializeTime` itself that several functions call this before running measurements.

As things stand, I'm inclined to remove this use of `initializeTime`, as it only serves to baffle users.

doomrobo: I agree that it couldn't hurt to mention that in the docs of `initializeTime`. The primary reasoning I have for including this line here is

1. I figure that many potential users of Criterion will learn from the provided examples and use the docs as necessary. At least that's how I start using a new library with a large API. I think it's conceivable that someone would see a call to `getTime` here and use it elsewhere without realizing the significance of `defaultMain`.

2. While I don't mind adding the information in the docs, I don't think users should have to know the implementation details of a function in order to see if a program is correct. That is to say, it wouldn't be obvious to me that this module is bug-free unless I already knew that `defaultMain` calls `initializeTime`.

3. The overhead that a single call to `initializeTime` incurs is negligible in almost any circumstance

4. Leaving out call to `initializeTime` results in behavior that's odd and hard to debug if you don't already know what you're looking for (as I have learned in the past week or so). So erring on the side of caution might save a lot of time for users.

So all things considered, I think having this here would be better for potential users of Criterion. If you still feel strongly against this, I'll remove it anyway and add the docs.

RyanGlScott: Fair enough.

I'm still not quite happy with the current state of the Haddocks for `initializeTime` and `getTime`, but I'll tackle that issue separately (see #197).

RyanGlScott:
Thanks, @doomrobo!

PR #200, opened 2018-08-08
Only depend on semigroups for GHC < 8.0.

quasicomputational:


quasicomputational:
Hmm. The Travis failure looks like it's falling over installing packages, before it's even looking at any Haskell code. Probably spurious and would, hopefully, go away with a restart.

Issue #202, opened 2018-09-15
Presence of '\'' in the benchmark's name breaks the html report


jhrcek:
When I use apostrophe character in benchmark's name, all the charts in the html report are broken.

Steps to reproduce
1.  use apostrophe in benchmark's name:

```haskell
module Main where

import Criterion
import Criterion.Main

main :: IO ()
main = defaultMain
    [ env (return ()) $
       \ ~() -> bgroup "Jan's benchmarks" [bench "dummy" $ nf id ()]
    ]
```
2. compile the file and run it, enabling html report generation:

./mybench --output report.html

3. Open the resulting report in the browser (I use latest google-chrom version 69)

Actual result:
none of the charts is displayed, and browser's console shows this error
`Uncaught SyntaxError: Unexpected identifier`
Inspecting the report more closely the error is coming from this piece of code generated in the report based on benchmark's name:
```javascript
 var ylabels = [[-0,'<a href="#b0">Jan's benchmarks/dummy</a>'],];
```
Literal inclusion of benchmark's name into the javascript apparently leads to premature finish of the string literal, causing the sytax error.

RyanGlScott:
Good catch. `criterion` uses the `microstache` library (a lighter version of `stache`) to substitute into `criterion`'s HTML template, and while `microstache` does correctly escape some characters, it seems to miss single quotes for some reason.

While #203 does fix this issue, I think a better fix would be to fix the upstream bug at `microstache`. I've submitted a bug report upstream at https://github.com/stackbuilders/stache/issues/35, and once that's fixed, I'll work on getting the fix into `microstache` itself.

RyanGlScott:
Fixed in #203.

PR #203, opened 2018-09-15
Fix broken html report when benchmark names contain '\''

jhrcek:
Proposed fix for https://github.com/bos/criterion/issues/202

### DIFF
--- a/templates/default.tpl
+++ b/templates/default.tpl
@@ -287,7 +287,7 @@ $(function () {
   reports.map(mangulate);
 
   var benches = [{{#report}}"{{name}}",{{/report}}];
-  var ylabels = [{{#report}}[-{{number}},'<a href="#b{{number}}">{{name}}</a>'],{{/report}}];
+  var ylabels = [{{#report}}[-{{number}},"<a href=\"#b{{number}}\">{{name}}</a>"],{{/report}}];
   var means = $.scaleTimes([{{#report}}{{anMean.estPoint}},{{/report}}]);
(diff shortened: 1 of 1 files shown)


jhrcek:
Hello @RyanGlScott
I changed the fix according to your suggestion and then investigate possibilities of writing automated test for it. I conclude that it would be too heavyweight to automate some kind of "interpret if the report is valid HTML + JS" just to verify this trivial fix.

Instead I just locally modified the `Sanity.hs` to generate reports for the first 10 000 characters 

```haskell
first10kChars = take 10000 [minBound::Char ..]
``` 

and then manually verified (by opening the report in the browser) which of these chars break the charts in the report, comparing the previous and fixed implementation.

Results:
The generated report is valid with all the tested chars, **except for* the following four exceptions :

| Dec| Character | Explanation |
|------|---------------|-----------------|
| 10 | \n | New line character breaks the template, as javascript string can't contain verbatim newline char|
| 13 | \r | Carriage return - same as above |
| 39 | ' | Apostrophe - **broke the previous implementation, but works fine with this PR** |
| 92 | \ | backslash - breaks the report only when at the end of the benchmark name, because it escapes the " that is supposed to end the string, but after escaping no longer does |

Even though it fixes just one char (and breaks no others) I'd suggest we merge this, because I feel it's relatively more likely that people include apostrophe in the benchmark name,

RyanGlScott:
Thanks for the thorough testing! I agree that it's probably best to just merge this as-is.

PR #209, opened 2019-01-29
Fix URL for benchmark screenshot

tchajed:
GitHub proxies all images (see their docs on [anonymized image URLs](https://help.github.com/articles/about-anonymized-image-urls/), and it seems www.serpentine.com isn't sending a Cache-Control: no-cache header for the screenshot and thus it isn't proxied correctly.

Fixed quite simply by using the screenshot in the repo; you can see this working at https://github.com/tchajed/criterion/tree/fix-readme-screenshot.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -5,7 +5,7 @@
 This package provides the Criterion module, a Haskell library for
 measuring and analysing software performance.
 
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Good catch—thanks for the patch.

PR #211, opened 2019-02-11
WIP: Regenerate .travis.yml

phadej:
Resolving https://github.com/haskell-CI/haskell-ci/issues/182
See PR https://github.com/haskell-CI/haskell-ci/pull/212

---

I reworked configuration of `haskell-ci`, so it's now easier to extend.


.travis.yml
```diff
@@ -146,16 +141,23 @@ script:
...
 
   # haddock
-  - if $HADDOCK; then cabal new-haddock -w ${HC} ${CABALFLAGS} ${TEST} ${BENCH} all; else echo "Skipping haddock generation";fi
+  - cabal new-haddock -w ${HC} ${TEST} ${BENCH} all
+
+  # Constraint sets
+  - rm -rf cabal.project.local
+
+  # Constraint set embed-files
+  - if [ $HCNUMVER -ge 80600 ] && [ $HCNUMVER -lt 80700 ] ; then cabal new-test  -w ${HC} --enable-tests  --enable-benchmarks  --constraint='criterion +embed-data-files' all ; fi
```
RyanGlScott: Shouldn't `--enable-tests  --enable-benchmarks` be `${TEST} ${BENCH}`?

cabal.haskell-ci
```diff
@@ -0,0 +1,10 @@
+ghc-head: True
+cabal-check: False
+no-tests-no-benchmarks: False
+unconstrained: False
+local-ghc-options: -Werror
+
+constraint-set embed-files
+  ghc: ==8.6.*
```
RyanGlScott: I'm impressed by the fact that `haskell-ci` can limit this constraint set to just GHC 8.6. That being said, the fact that I did so in the original version was a crude hack. I'd actually want to run this on //all// versions of GHC that are tested here. (In other words, you can delete this line.)

.travis.yml
RyanGlScott: Also, the mechanism for accomplishing this seems a bit crude. I would have expected something closer to this:

```bash
do
  cabal new-build -w ${HC} ${TEST} ${BENCH} --constraint='criterion +embed-data-files' all
  cabal new-test  -w ${HC} ${TEST} ${BENCH} --constraint='criterion +embed-data-files' all
done
```

Not only does this more closely mirror what we're doing in the non-`embed-data-files` case, but it would also make it easier to add or drop lines as necessary. (For instance, I would have hoped that we could run the Haddocks with `embed-data-files` enabled as well.)

phadej: I'll go with this for now, feel free to make a follow-up PR in `haskell-ci`.

RyanGlScott: I mean... you're still actively working on haskell-CI/haskell-ci#212, right? It can't be that hard to just add this from the get-go.

cabal.haskell-ci
phadej: FWIW, for now the `ghc: ` field as required, but it can be changed to optional with "all ghc" as the default.

.travis.yml
phadej: I spent most of the day (which is not work time) on it. Sorry, have to draw a line somewhere.

RyanGlScott: Sure, that's fine. But do be aware that I'm not going to merge this until we can reach feature parity with what the current `.travis.yml` does.

phadej: This PR is a demo, to check what travis says :) you can can close this after I get green from travis.

RyanGlScott: No worries! I do want to land something analogous to this at some point, so I'd be fine with keeping this open for testing further changes. (I'll mark it as a WIP in the meantime.)

phadej:
This is green. I'm happy for now.

RyanGlScott:
Commits c4f062c0ea36796b5d1a0a8333ff45abead50c85 and d4cb19dfcce726388761b86a93a4356107b9fca0 are the spiritual successor to this PR.

PR #216, opened 2019-08-14
WIP: Support GHCJS

phadej:
Depends on
- https://github.com/bos/math-functions/pull/55
- https://github.com/bos/statistics/pull/153 or/and https://github.com/bos/statistics/pull/154

A "proof" that it works:

Screenshot from 2019-08-15 01-00-56

(and is 20x or so slower than GHC, for what it's worth)

### DIFF
--- a/Criterion/Monad.hs
+++ b/Criterion/Monad.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE Trustworthy #-}
 -- |
 -- Module      : Criterion.Monad
@@ -22,7 +23,13 @@ import Criterion.Monad.Internal (Criterion(..), Crit(..))
 import Criterion.Types hiding (measure)
 import Data.IORef (IORef, newIORef, readIORef, writeIORef)
 import System.IO.CodePage (withCP65001)
-import System.Random.MWC (GenIO, createSystemRandom)
+import System.Random.MWC (GenIO
+#if defined(__GHCJS__)
+  , create
+#else
+  , createSystemRandom
+#endif
+  )
 
 -- | Run a 'Criterion' action with the given 'Config'.
 withConfig :: Config -> Criterion a -> IO a
@@ -35,7 +42,12 @@ withConfig cfg (Criterion act) = withCP65001 $ do
 -- This is not currently thread-safe, but in a harmless way (we might
 -- call 'createSystemRandom' more than once if multiple threads race).
 getGen :: Criterion GenIO
-getGen = memoise gen createSystemRandom
+getGen = memoise gen
+#if defined(__GHCJS__)
+  create -- there is no /dev/urandom or GHCJS, let's just use fixed seed.
+#else
+  createSystemRandom
+#endif
 
 -- | Memoise the result of an 'IO' action.
 --
--- a/criterion-measurement/src/Criterion/Measurement.hs
+++ b/criterion-measurement/src/Criterion/Measurement.hs
@@ -49,18 +49,29 @@ import GHC.Stats (GCStats(..))
 #endif
 import Prelude ()
 import Prelude.Compat
-#if MIN_VERSION_base(4,7,0)
-import System.Mem (performGC, performMinorGC)
-# else
-import System.Mem (performGC)
-#endif
+
 import Text.Printf (printf)
 import qualified Control.Exception as Exc
 import qualified Data.Vector as V
 import qualified GHC.Stats as Stats
 
-#if !(MIN_VERSION_base(4,7,0))
+-- https://github.com/ghcjs/ghcjs-boot/issues/32
+#if defined(__GHCJS__)
+import System.Mem (performGC)
+
+performMinorGC :: IO ()
+performMinorGC = return ()
+#else
+
+# if MIN_VERSION_base(4,7,0)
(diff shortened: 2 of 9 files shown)


criterion.cabal
```diff
@@ -100,7 +101,7 @@ library
     mwc-random >= 0.8.0.3,
     optparse-applicative >= 0.13,
     parsec >= 3.1.0,
-    statistics >= 0.14 && < 0.16,
+    statistics >= 0.14 && < 0.17,
```
RyanGlScott: Why the major version bump here?

Criterion/Monad.hs
```diff
@@ -35,7 +42,12 @@ withConfig cfg (Criterion act) = withCP65001 $ do
 -- This is not currently thread-safe, but in a harmless way (we might
 -- call 'createSystemRandom' more than once if multiple threads race).
 getGen :: Criterion GenIO
-getGen = memoise gen createSystemRandom
+getGen = memoise gen
+#if defined(__GHCJS__)
+  create -- there is no /dev/urandom or GHCJS, let's just use fixed seed.
```
RyanGlScott: Is there really no other source of randomness we could use here?

criterion-measurement/src/Criterion/Measurement.hs
```diff
@@ -119,7 +130,10 @@ data GCStatistics = GCStatistics
 --
 -- If you need guaranteed up-to-date stats, call 'performGC' first.
 getGCStatistics :: IO (Maybe GCStatistics)
-#if MIN_VERSION_base(4,10,0)
+#if defined(__GHCJS__)
+getGCStatistics = return Nothing
```
RyanGlScott: Is there a GHCJS-specific reason why this can't be done? If so, can you leave a comment explaining why not?

criterion-measurement/src/Criterion/Measurement.hs
```diff
@@ -390,6 +404,31 @@ secs k
                | t >= 1e1  = printf "%.2f %s" t u
                | otherwise = printf "%.3f %s" t u
 
+#if defined(__GHCJS__)
+-- atm, cabal haddock --ghcjs fails :(
```
RyanGlScott: Is this due to haskell/cabal#5840? If so, can you leave a link to it in the comments?

criterion-measurement/src/Criterion/Measurement.hs
```diff
@@ -390,6 +404,31 @@ secs k
...
+getCPUTime = return 0
+
+-- Performance API is in all major browsers: https://developer.mozilla.org/en-US/docs/Web/API/Performance
+-- and in NodeJS since 8.5.0: https://nodejs.org/api/perf_hooks.html#perf_hooks_performance_timing_api
+--
+-- Examples of GHCJS FFI: https://github.com/ghcjs/ghcjs/wiki/A-few-examples-of-Foreign-Function-Interface
+--
+-- performance.now() returns *milliseconds*. So we divide them to get seconds.
+foreign import javascript unsafe
+  "if (typeof window === 'object' && typeof window.performance === 'object') { $r = window.performance.now() / 1000; } else if (typeof require === 'function') { $r = require('perf_hooks').performance.now() / 1000; } else { $r = 0.0; }"
```
RyanGlScott: This is... a lot of JavaScript to put on a single line. Is there any way to factor this out somehow?

examples/Overhead.hs
```diff
@@ -12,7 +12,11 @@ import GHC.Stats as GHC
 main :: IO ()
 main = do
   M.initializeTime -- Need to do this before calling M.getTime
+#if defined(__GHCJS__)
+  statsEnabled <- return False
```
RyanGlScott: Instead of adding CPP here, change the definition of the `getRTSStatsEnabled` backwards-compatibility shim on line 36 of this file.

(Also, if there is a GHCJS-specific reason why this must return `False`, it would be good to leave a comment explaining why.)

criterion.cabal
phadej: Indeed, it isn't necessary as `statistics` worked around GHCJS limitations without major bump.

Criterion/Monad.hs
phadej: One can play with `time`, I'm too lazy.

criterion-measurement/src/Criterion/Measurement.hs
phadej: Yes, IIRC, the C-primitives to lookup RTS stats are simply not implemented in GHCJS C-shims.

criterion-measurement/src/Criterion/Measurement.hs
phadej: yes, that's why I didn't bother to haddock those properly (though looks like I did something anyway.)

criterion-measurement/src/Criterion/Measurement.hs
phadej: In separate file? I don't know how, maybe. If we had multiline string literals, it would be ok; it's not that long after all.
Comment above hints what happens there.

examples/Overhead.hs
phadej: done

phadej:
I made some changes, but in general lost interest in getting this done. If someone needs this too, feel free to continue.

PR #221, opened 2020-02-21
Criterion-measurements with even less dependencies

hsyl20:
Remove dependencies on `vector` and `aeson`. Allows `criterion-measurements` to be used only with boot libraries (and `base-compat`)

### DIFF
--- a/Criterion/Analysis.hs
+++ b/Criterion/Analysis.hs
@@ -140,21 +140,23 @@ analyseSample :: Int            -- ^ Experiment number.
               -> String         -- ^ Experiment name.
               -> V.Vector Measured -- ^ Sample data.
               -> ExceptT String Criterion Report
-analyseSample i name meas = do
+analyseSample i name measV = do
   Config{..} <- ask
   let ests      = [Mean,StdDev]
+      measL     = V.toList measV
       -- The use of filter here throws away very-low-quality
       -- measurements when bootstrapping the mean and standard
       -- deviations.  Without this, the numbers look nonsensical when
       -- very brief actions are measured.
-      stime     = measure (measTime . rescale) .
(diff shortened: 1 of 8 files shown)


RyanGlScott:
Thanks for the pull request.

While I can understand the desire to minimize dependencies, I'm not sure if the benefits outweigh the drawbacks in this case. Specifically:

1. The `FromJSON`/`ToJSON` instances for `Measured` are now orphans. In general, I'm very reluctant to introduce orphan reasons without a very compelling reason.
2. I'm worried that removing the use of `vector` in `criterion-measurement` will impact the performance characteristics of the functions that previously used vector types. In particular, there are now several intermediate `fromList`/`toList` conversions in `criterion` to fix the impedance mismatches between lists and vectors. These conversions can add up, and I don't think minimizing dependencies is a good reason to justify this.

hsyl20:
The rationale for this was to benchmark GHC HEAD (building `aeson` and `vector` with it is a pain). I had this code in a branch since last year and I figured I would share it upstream as I needed it again today. If the drawbacks outweigh the benefits I can make another package out of this.

RyanGlScott:
Coincidentally enough, I build `criterion` with GHC HEAD all the time. I've never had any particular difficulties in doing so either due to [`head.hackage`](https://gitlab.haskell.org/ghc/head.hackage), but perhaps I'm not doing the same things with HEAD that you are.

If no-dependencies-besides-boot-libraries is a must, you might also consider [`gauge`](http://hackage.haskell.org/package/gauge), which is a fork of `criterion` that minimizes dependencies as far as possible.

hsyl20:
I have been reluctant to use/setup `head.hackage`. It seemed easier to remove the dependencies from `criterion-measurement` than to try to use it. ;-) Thanks for the link to `gauge`.

Issue #207, opened 2018-12-18
Legend boxes can hide data points in HTML file


nh2:
This can be quite confusing; pictures below:

roman:
<img width="530" alt="screenshot 2018-12-18 11 46 17" src="https://user-images.githubusercontent.com/7335/50178996-ce4f8000-02ba-11e9-811c-08a6729c49ad.png">
<img width="523" alt="screenshot 2018-12-18 11 46 38" src="https://user-images.githubusercontent.com/7335/50178997-ce4f8000-02ba-11e9-88b5-c26c49a38018.png">

nh2:
There is already some transparency on the white `<div>` but depending on the screen it is _extremely_ hard to see.

We should either increase the transparency, or move the legend outside of the plot (e.g. further up).

Lythenas:
I'm also having this Problem with the overview chart when using long names. Which basically makes the chart unreadable.

![image](https://user-images.githubusercontent.com/32217236/63586605-f574f480-c5a1-11e9-90c6-ed284fad2163.png)

Also the legend can be bigger than the actual graph.

![image](https://user-images.githubusercontent.com/32217236/63586643-15a4b380-c5a2-11e9-9ed3-875b60a773a3.png)

I think moving the legend above the plot would be fine. Except maybe for the overview chart. There is also more than enough space on the left and right to move the legend there. But that probably depends on screen resolution.

Maybe something like transparency (or even completely invisible) on hover would also work.

RyanGlScott:
#229 changes the way legends work:

* Legends are no longer displayed over the time-per-iteration charts, as they don't really convey much useful information beyond what is already present in the titles. This should resolve the issue in https://github.com/bos/criterion/issues/207#issuecomment-448346753.
* By default, legends are no longer displayed on top of the overview chart, which should fix https://github.com/bos/criterion/issues/207#issuecomment-524266321. The legend can be toggled, however, and if it is enabled, it is no longer rendered on top of the bars themselves.

As a result, I'm deeming this issue fixed. Please re-open if you disagree.

PR #229, opened 2020-11-06
Port HTML report to chart.js

### LINKED ISSUE #213: Logarithmic scale when plotting results
Hi! 
I was wondering if there was such a feature in criterion, and if not how would I go about implementing it in the library?

jonascarpay:
Originally intended to close #213, has turned into a rewrite using `chart.js`.

[`fibber.html`](https://github.com/jonascarpay/criterion/blob/master/www/fibber.html)
[`report.html`](https://github.com/jonascarpay/criterion/blob/master/www/report.html)

Feedback, criticism, feature requests welcome!
### To do
- [x] Report/template
  - [x] better error bars
  - [x] Colophon
  - [x] Explanation
  - [x] Outlier info
  - [x] test with different data sets
  - [x] ~(optional?) legends~
  - [x] strut 0-width bars in overview
  - [x] fix tooltip on scatter plot
  - [x] ~(collapsible) table of contents?~
  - [x] general code cleanup
  - [x] git history cleanup
  - [x] make titles links
  - [x] things are too bold
  - [x] link style more subtle
  - [x] overview fixed height, responsive width
  - [x] stack KDE and scatter at small widths
  - [x] make tick display more robust
  - [x] move example chart to www/report.html
  - [x] match precision with old report
  - [x] change cursor when hovering y-axis
  - [x] change sort order
    - [x] interface?
    - [x] report order
    - [x] alphabetical order
    - [x] lacitebahpla order
    - [x] time order

- [x] Haskell+packaging
  - [x] Drop `js-flot`, `js-jquery`
  - [x] Include `chart.js`
    - [x] make `js-chart`?
    - [x] EMBED flag
  - [x] split out custom JS+CSS?
    
@considerate @cartazio


RyanGlScott:
Hooray! It looks like this is still a draft, so let me know when this is ready for a full review.

jonascarpay:
This is ready for review. The fact that it's one commit is a bit unfortunate, but since most work happened in files that have since disappeared there is not really a nice way to make it more atomic.

RyanGlScott:
Awesome! I'll do a more thorough review of this patch over the weekend, but a quick before-and-after comparison makes me notice two things right off the bat:

* The legends appear to be missing after your changes. Is this intentional?
* After your changes, it is no longer possible to manually zoom in/out or pan horizontally on the overview chart. I'm guessing this is because this functionality has been replaced by the ability to click on an individual benchmark and have the chart automatically zoom in/out accordingly? If so, perhaps that's a reasonable justification for removing this functionality, although I wonder if certain people relied on this.

Some other random thoughts:

* There are some not-so-obviously discoverable features in the new chart engine, such as the ability to toggle between linear and log scales by clicking on the _x_-axis. I wonder if it would be worth describing these features in prose somewhere on the HTML page itself?
* I'm also not clear on what some of the new drop-down menu options (e.g., colex) mean. These might be worth spelling out.
* I'm not sure if it's just my web browser, but trying to hover my mouse over the yellow trendlines in the per-benchmark reports feels a _lot_ more finicky after your changes, in the sense that the timing information in a tooltip won't appear unless my mouse is directly over the line. In contrast, with today's `criterion`, your mouse doesn't have to be directly over the trendline in order for the tooltip to appear; as long as it is close enough to the trendline, the tooltip will surface. Is it possible to increase the "hitbox" of the trendline so that it mimics the behavior of today's `criterion`?

jonascarpay:
> The legends appear to be missing after your changes. Is this intentional?

This was mostly because they seemed to duplicate information already present on the same screen. It's fairly trivial to add them back in, would you like me to?

> After your changes, it is no longer possible to manually zoom in/out or pan horizontally on the overview chart.

Yeah, zooming was left out intentionally. I figured the primary use case was to see bars that got dwarfed by others, which I felt the click-to-focus was a better interface for. It's possible to do zooming, but it also requires a disproportionate amount of code to get right. What do you think?

> There are some not-so-obviously discoverable features in the new chart engine, such as the ability to toggle between linear and log scales by clicking on the x-axis.

Makes sense, I added a note.

> I'm also not clear on what some of the new drop-down menu options (e.g., colex) mean. These might be worth spelling out.

Both lexical and colex sort by group, with lexical the head is most significant and with [colex the tail is most significant](https://en.wikipedia.org/wiki/Lexicographic_order#Colexicographic_order) (i.e. big-endian vs little-endian). I suppose "by group" would be a better name for lexical, do you know a good, self-evident name for colex?

> trying to hover my mouse over the yellow trendlines in the per-benchmark reports feels a lot more finicky after your changes

414911b expanded the radius, do you think it's OK now?

considerate:
@RyanGlScott Note that `report.html` haven’t been regenerated with the updated radius for the hit boxes yet so to test the tooltips you would have to generate a new report.

considerate:
@jonascarpay @RyanGlScott I added titles to the kernel density estimates and scatter plots to provide the same amount of context as the previous version for those plots.

I think it might somewhat redundant and would personally like to keep the output as free from noise as possible. What do you think?

RyanGlScott (commented):
Thanks for looking into the legend issue. I'm now reminded of the existence of #207, however, where people specifically complained that legends were making their charts unreadable. In light of this, I'm wondering if I should take back my previous suggestion to reinstate the titles. Or, at the very least, allow one to toggle whether the legend appears or not with a mouse click. What do you think?

> Yeah, zooming was left out intentionally. I figured the primary use case was to see bars that got dwarfed by others, which I felt the click-to-focus was a better interface for. It's possible to do zooming, but it also requires a disproportionate amount of code to get right. What do you think?

Upon further thought, I agree with your assessment that click-to-zoom is a better interface for this.

> Both lexical and colex sort by group, with lexical the head is most significant and with [colex the tail is most significant](https://en.wikipedia.org/wiki/Lexicographic_order#Colexicographic_order) (i.e. big-endian vs little-endian). I suppose "by group" would be a better name for lexical, do you know a good, self-evident name for colex?

Ah, "colex" stands for "colexical"! That's illuminating, especially since I couldn't even find what colex meant in a search engine. At the very least, I would recommend expanding "colex" to "colexical". I would also recommend spelling out what each of these options mean in the "want to understand this report?" section.

> 414911b expanded the radius, do you think it's OK now?

Much better, thanks!

> I added titles to the kernel density estimates and scatter plots to provide the same amount of context as the previous version for those plots.
> 
> I think it might somewhat redundant and would personally like to keep the output as free from noise as possible. What do you think?

Personally, I'm a fan of the titles and labels, since it makes the information more self-contained. On the other hand, the y-axis labels do cause the actual charts themselves to shrink a fair bit in order to accommodate the new text, so I can definitely understand your reservations there. Ultimately, I think this comes down to a judgment call about optimizing the amount of information displayed versus the size of the chart. Given that the axis labels are implied by the titles themselves, I think I would be fine with omitting them.

RyanGlScott:
I've timed out for now, but this is looking very promising. More review to come later!

cartazio:
Is there a preview link that works on mobile? :))

jonascarpay:
Alright, that should do it for the feedback up to now.

@cartazio Refined Github gives me [this](https://refined-github-html-preview.kidonng.workers.dev/jonascarpay/criterion/raw/master/www/report.html) preview link, I think that should work. It does look like on mobile part of the overview chart gets eaten, will look into that.

cartazio:
Hrmm. One good thing to test is having an example plot where there’s a huge
range of magnitudes of values.

Def like how this one looks so far.

One issue with the previous was some sort of huge algorithmic penalty for
actually rendering some of the plots. Was that part of the issues with the
old flot code?

jonascarpay:
> One good thing to test is having an example plot where there’s a huge
range of magnitudes of values.

[This is the benchmark suite we've used for that purpose before](https://refined-github-html-preview.kidonng.workers.dev/jonascarpay/criterion/raw/example_report/report.html)

> One issue with the previous was some sort of huge algorithmic penalty for actually rendering some of the plots.

Hmm, I don't remember there being any such algorithm in the old code. Do you remember how that issue would manifest? 

> Was that part of the issues with the old flot code?

No, the issue was a choice between a) the stable version that didn't support things like log axes and b) the new version that was unstable in every sense of the word

considerate:
> One issue with the previous was some sort of huge algorithmic penalty for
actually rendering some of the plots. Was that part of the issues with the
old flot code?

Without running any proper benchmarks (ironic, isn't it?) generating the same charts took much longer with flot than with chart.js. That said, it still takes a non-trivial amount of time to render the benchmark link posted above on mobile.

templates/default.tpl
```diff
@@ -1,362 +1,138 @@
...
+        <li><em>the chevron</em> in the top-right toggles the the legend</li>
+        <li><em>a group name in the legend</em> shows/hides that group</li>
+      </ul>
+    </p>
+
+    <p>
+      The overview chart supports the following sort orders:
+      <ul>
+        <li><em>index</em> order is the order as the benchmarks are defined in criterion</li>
+        <li><em>lexical</em> order sorts groups left-to-right, alphabetically</li>
```
RyanGlScott, line 57: Shouldn't it be "top-to-bottom", not "left-to-right", since the bars are arranged vertically?

templates/default.tpl
```diff
@@ -1,362 +1,138 @@
...
+        <li><em>colexical</em> order sorts groups right-to-left, alphabetically</li>
```
RyanGlScott, line 58: Similarly, shouldn't this be "bottom-to-top", not "right-to-left"?

Criterion/Report.hs
```diff
@@ -128,11 +123,8 @@ formatReport reports templateName = do
     let context = object
             [ "json"                .= reports
             , "report"              .= reports'
```
RyanGlScott, line 125: I don't think the new templates make use of `report`, right? If so, can we remove this line?

templates/default.tpl
jonascarpay: The left-to-right refers to that the groups are sorted with the leftmost being the most significant. Can you think of a better way to word it?

Criterion/Report.hs
jonascarpay: Fixed in d381808

Criterion/Report.hs
```diff
@@ -185,58 +170,6 @@ formatReport reports templateName = do
...
-        times  = measure measTime reportMeasured
-        cycles = measure measCycles reportMeasured
-        anMeanConfidenceLevel
-               = confidenceLevel $ confIntCL $ estError anMean
-        (anMeanLowerBound, anMeanUpperBound)
-               = confidenceInterval anMean
-        (anStdDevLowerBound, anStdDevUpperBound)
-               = confidenceInterval anStdDev
-
-        sanitizeJSString :: T.Text -> IO T.Text
```
RyanGlScott, line 229: The removal of `sanitizeJSString` makes me worried that we're accidentally reverting the fix for #224. How does this branch render the program in https://github.com/bos/criterion/issues/224#issue-611029139?

templates/criterion.js
```diff
@@ -0,0 +1,870 @@
...
+    var overviewLineHeight = 16 * 1.25;
+    overview.style.height =
+      String(overviewLineHeight * reportData.length + 36) + 'px';
+    overview.appendChild(mkOverview(reportData.slice()));
+    var reports = document.getElementById('reports');
+    reportData.forEach(function(report, i) {
+      var id = 'report_' + String(i);
+      reports.appendChild(
+        elem('div', {id: id, className: 'report-details'}, [
+          elem('h1', {}, [elem('a', {href: '#' + id}, [report.groups.join(' / ')])]),
```
RyanGlScott, line 863: Does this handle apostrophes in group names correctly? (See #202.)

templates/default.tpl
RyanGlScott: I see what you mean, although I still think the status quo is a bit confusing. Ascending/descending order is perhaps the least ambiguous way of referring to the property of being sorted a certain way, so I would phrase this as:

> _lexical_ order sorts groups in ascending alphabetical order

templates/default.tpl
RyanGlScott: > _colexical_ order sorts groups in descending alphabetical order

templates/criterion.js
considerate: This creates a [text node](https://developer.mozilla.org/en-US/docs/Web/API/Document/createTextNode), which escapes HTML. However, it doesn't style the apostrophes in any specific way (by replacing with with i.e. a [right single quote](http://www.fileformat.info/info/unicode/char/2019/index.htm)). The issue in #202 should not be a problem here though.

Criterion/Report.hs
considerate: We're not injecting any JavaScript any more. Instead a `data-report-json` attribute is used in an HTML tag which is then read and parsed with `JSON.parse`. This is precisely to avoid sanitizing and escaping javascript manually which is really tricky to get right.

considerate:
> > colexical order sorts groups in descending alphabetical order

I'm not sure I agree that this description is correct, the groups are sorted in ascending alphabetical order but we split the report name on '/' and compare the two right-most group elements first with  (the default) ascending alphabetical string comparison, then if equal the group elements to the left of those are compared until one of the elements are unequal. If all elements are equal the shorter array is sorted before the longer one.

templates/criterion.js
```diff
@@ -0,0 +1,870 @@
...
+    }
+    return aParts.length - bParts.length;
+  }
+  function lexicalSort(a, b) {
+    return lex(a.groups, b.groups);
+  }
+
+  function reverseLexicalSort(a, b) {
+    return lex(a.groups.slice().reverse(), b.groups.slice().reverse());
+  }
```
considerate, line 161: This is the implementation for the `lexical` and `colexical` sort orders. The implementation is the same, except that the colexical first reverses the two group arrays performing a normal `lexical` sort.

templates/default.tpl
```diff
@@ -1,362 +1,138 @@
...
+      <span class="overview-info">
+        <a href="#controls-explanation" class="info" title="click bar/label to zoom; x-axis to toggle logarithmic scale; background to reset">&#9432;</a>
+        <a id="legend-toggle" class="chevron button"></a>
+      </span>
+    </div>
+    <aside id="overview-chart"></aside>
+    <main id="reports"></main>
 
-   </div>
+    <div id="report-data" data-report-json='{{{json}}}'></div>
```
considerate, line 36: The report JSON data is added here.

templates/criterion.js
```diff
@@ -0,0 +1,870 @@
...
+      }))
+    ]);
+  }
+  document.addEventListener('DOMContentLoaded', function() {
+    var reportData = JSON.parse(document.getElementById('report-data')
+      .getAttribute('data-report-json'))
+      .map(function(report) {
+        report.groups = report.reportName.split('/');
+        return report;
+      });
```
considerate, line 851: The report data is parsed here.

RyanGlScott:
> [This image](https://en.wikipedia.org/wiki/Lexicographic_order#/media/File:Orderings;_permutations_5-cycle.svg) provides a visualization of the lexical and colexical orderings. The right-most 1s are placed before the right-most 5s in colexical order.

OK. My attempts at succinctly characterizing this have failed, so perhaps it's best that I stop trying. :)

In that case, I'm content with the current description, but include a link to the Wikipedia article on lexicographic ordering you mentioned above, since that would be valuable context to have.

RyanGlScott:
One last request: do you mind writing up a description of the changes from this patch in the `changelog.md` file? There have been quite a few changes (both in terms of new functionality and UI differences), and it would be good to advertise them prominently. Thanks!

considerate:
> One last request: do you mind writing up a description of the changes from this patch in the changelog.md file? There have been quite a few changes (both in terms of new functionality and UI differences), and it would be good to advertise them prominently. Thanks!

Yes, of course, I'll co-ordinate with @jonascarpay and describe the changes in the changelog either today or tomorrow.

jonascarpay:
Alright, that's the Changelog. Do you want me to squash the history?

RyanGlScott:
> Alright, that's the Changelog. Do you want me to squash the history?

Please do. And thanks again for all of your combined work on this!

jonascarpay:
Done, thank you as well!

cartazio:
I’m so excited about this. Thank you all so much!

RyanGlScott:
I've uploaded [`criterion-1.5.8.0`](http://hackage.haskell.org/package/criterion-1.5.8.0) to Hackage with these changes.

PR #232, opened 2020-11-16
Escape JSON to avoid parsing the content as HTML

considerate:
By crafting a malicious string it was quite easy to escape the HTML
attribute that contained the JSON report data allowing injection of
arbitrary HTML using the report names.

This commit aims to mitigate the breakage of the reports when report names
contain unexpected characters. The JSON data is moved to a <script> tag and
the content is escaped by replacing significant HTML characters with their
corresponding JSON escape sequences. The '<' character is replaced with the
sequence "\u003c", '&' is replaced with "\u0026" and so on.

screenshot-2020-11-16T23:03:20+09:00

The processing of `/` is still not ideal, if a report name contains that character
then it is treated as a group separator.

### DIFF
--- a/Criterion/Report.hs
+++ b/Criterion/Report.hs
@@ -32,12 +32,13 @@ module Criterion.Report
     ) where
 
 import Control.Exception (Exception, IOException, throwIO)
-import Control.Monad (mplus, unless)
+import Control.Monad (mplus)
 import Control.Monad.IO.Class (MonadIO(liftIO))
 import Control.Monad.Reader (ask)
 import Criterion.Monad (Criterion)
 import Criterion.Types
 import Data.Aeson (ToJSON (..), Value(..), object, (.=), Value)
+import Data.Aeson.Text (encodeToLazyText)
 import Data.Data (Data, Typeable)
 import Data.Foldable (forM_)
 import GHC.Generics (Generic)
@@ -46,8 +47,8 @@ import Statistics.Function (minMax)
 import System.Directory (doesFileExist)
 import System.FilePath ((</>), (<.>), isPathSeparator)
 import System.IO (hPutStrLn, stderr)
-import Text.Microstache (Key (..), MustacheWarning (..), Node (..), Template (..),
-                         compileMustacheText, displayMustacheWarning, renderMustacheW)
+import Text.Microstache (Key (..), Node (..), Template (..),
+                compileMustacheText, displayMustacheWarning, renderMustacheW)
 import Prelude ()
 import Prelude.Compat
 import qualified Control.Exception as E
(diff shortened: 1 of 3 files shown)


Criterion/Report.hs
```diff
@@ -103,6 +104,22 @@ report reports = do
     tpl <- loadTemplate [td,"."] template
     TL.writeFile name =<< formatReport reports tpl
 
+-- | Escape JSON string aimed to be embedded in an HTML <script> tag.
+-- Notably < and > are replaced with their unicode escape sequences such
+-- that closing the <script> tag from within the JSON data is disallowed.
+escapeJSON :: Char -> TL.Text
```
RyanGlScott, line 110: Out of curiosity, where did you get the implementation for this function? I've seen various iterations of this function before, most notably in GHCJS ([here](https://github.com/ghcjs/ghcjs/blob/718b37fc7167269ebca633914c716c3dbe6d0faf/src/Compiler/JMacro/Base.hs#L887-L905)). The GHCJS version has quite a few differences from the code shown here, and I wonder if these are important.

Criterion/Report.hs
```diff
@@ -120,8 +137,10 @@ formatReport reports templateName = do
     templates <- getTemplateDir
     template <- includeTemplate (includeFile [templates]) template0
 
+
+    let reportJSON = TL.toStrict . TL.concatMap escapeJSON . encodeToLazyText
     let context = object
-            [ "json"                .= reports
+            [ "json"                .= String (reportJSON reports)
```
RyanGlScott, line 143: Is the use of `String` here needed? The `ToJSON` instance for `Text` should take care of this, if I recall correctly.

Criterion/Report.hs
```diff
@@ -120,8 +137,10 @@ formatReport reports templateName = do
     templates <- getTemplateDir
     template <- includeTemplate (includeFile [templates]) template0
 
+
+    let reportJSON = TL.toStrict . TL.concatMap escapeJSON . encodeToLazyText
```
RyanGlScott, line 141: Minor stylistic nitpick: move `reportJSON` to the `where` clause, as its definition does not depend on anything defined in the `do` block.

RyanGlScott (commented):
Good catch! Is this bug present in all versions of `criterion`, or only after #229?

Criterion/Report.hs
considerate: It might not be, I started out with `ByteString` and got type errors. I’ll see if it works without the wrapper.

Criterion/Report.hs
considerate: The GHCJS escaping seems to escape the JSON with respect to other JSON (or maybe JavaScript), but I'm trying to escape the JSON with respect to HTML here.

I started out with a normal XML escape using [`&<control>;` character references](https://html.spec.whatwg.org/multipage/syntax.html#character-references) from the provided mustache library, escaping the content in the attribute of the div tag originally used for embedding the JSON data. However, with this approach I was able to exit the quotes wrapping the JSON data and inject .

Then I tried to escape this attribute by writing a small escape sequence for the strings but given that both string and HTML escaping was necessary I went with the approach of just directly embedding the data in a <script> tag.

With this approach I once again tried with HTML escaping but this time those escape sequences were left intact after reading the data from the DOM. Thus I turned to escaping with [JSON unicode sequences](https://tools.ietf.org/html/rfc8259#section-7) instead. The most obvious character to escape is `<` to avoid parsing [end tags](https://html.spec.whatwg.org/multipage/syntax.html#end-tags) (more specifically `<script>`), however when looking into the problem I found that using `+` and [UTF-7 escape codes](https://en.wikipedia.org/wiki/UTF-7#Security) you can also trick some browsers to render a script tag. The unicode escape sequence approach is safe to use since after applying `JSON.parse` you get back exactly the code point the sequence resolves to which is the same character as we started with in Haskell. However, since HTML doesn't resolve these sequences and the `<script>` tag is listed as [raw text](https://html.spec.whatwg.org/multipage/syntax.html#raw-text-elements), it uses only uses ampersand for escaping there's no issue with the HTML parser treating the JSON as HTML data.

In conclusion we only _really_ need to escape `<`, `&`, and `+` inside a `<script>` tag for a standards compliant browser but given all inconsistencies in implementations I figured it might be wise to escape a bit more than strictly necessary. Things like null characters trip up a lot of implementations in weird places. Note that I didn't quote `"` since that would unnecessarily result in a lot of extra bytes in the output.

considerate: I found a link to a Python library containing a helper function with the same goal: 

https://github.com/pallets/jinja/blob/9242a027c926a4636d33bac95b1f7654d8003ac9/src/jinja2/utils.py#L539

considerate: This is what `Rails` does in their `json_escape` function:
https://github.com/rails/rails/blob/1165401ee962aef0aaf81080e3e7dcab522efa40/activesupport/lib/active_support/core_ext/string/output_safety.rb#L10
https://github.com/rails/rails/blob/1165401ee962aef0aaf81080e3e7dcab522efa40/activesupport/lib/active_support/core_ext/string/output_safety.rb#L113

Criterion/Report.hs
```diff
@@ -131,18 +150,11 @@ formatReport reports templateName = do
...
-      --
-      -- So `json` represents a raw JavaScript array. This is a bit skeevy by
-      -- mustache conventions, but redesigning the template to avoid this
-      -- warning would be more work than just substituting the array directly.
-      unless (warning == MustacheDirectlyRenderedValue (Key ["json"])) $
         criterionWarning $ displayMustacheWarning warning
     return formatted
   where
+    reportsJSON :: [Report] -> T.Text
+    reportsJSON = TL.toStrict . TL.concatMap escapeJSON . encodeToLazyText
```
RyanGlScott, line 157: Nitpick: put a newline after `reportJSON` for consistency with other functions defined in this `where` clause.

Criterion/Report.hs
RyanGlScott: > The GHCJS escaping seems to escape the JSON with respect to other JSON (or maybe JavaScript), but I'm trying to escape the JSON with respect to HTML here.

Ah, that's an important distinction that I didn't realize.

> In conclusion we only _really_ need to escape `<`, `&`, and `+` inside a `<script>` tag for a standards compliant browser but given all inconsistencies in implementations I figured it might be wise to escape a bit more than strictly necessary. Things like null characters trip up a lot of implementations in weird places. Note that I didn't quote `"` since that would unnecessarily result in a lot of extra bytes in the output.

Gotcha. Since this part wasn't obvious to me until you pointed it out, could you include some version of this in the Haddocks/comments?

RyanGlScott (approved):

RyanGlScott:
Thanks again! I've uploaded [`criterion-1.5.9.0`](http://hackage.haskell.org/package/criterion-1.5.9.0) to Hackage with these changes.

considerate:
@RyanGlScott Thank you!

PR #234, opened 2021-02-28
Add GHC-9.0 job to CI

phadej:


RyanGlScott:
Thanks! Out of curiosity, can we expect a new Hackage release of `haskell-ci` soon? Admittedly, I was waiting on that before I personally went through and updated all of the code I maintained to test 9.0.1 on CI.

phadej:
@RyanGlScott I want to at least try to add GHCJS support for GHA before doing the release. In March.

PR #236, opened 2021-03-25
Use a more precise type for nfIO'

ckoparkar:
Its first argument is supposed to be a function which reduces the result to normal form. Specifying its type as 'a -> ()', rather than 'a -> b' conveys this intention better. Also, presently nfIO' is an internal function used only by nfIO, which uses 'rnf' as a reducer.

Anyway, this is a really minor thing. I was looking at this file for something and happened to notice this.


RyanGlScott:
Good catch! `nfAppIO'` already uses `()` as the return type for its `reduce` function, but for whatever reason, `nfIO'` wasn't doing the same. Let's fix that.

PR #247, opened 2021-10-09
Fix #246: Support aeson-2.0.0.0

### LINKED ISSUE #246: Support aeson-2
The latest criterion fails to build with it.

I'll submit patch shortly, as well as make the appropriate revisions. (my machine is busy verifying them atm).

phadej:


Criterion/Report.hs
```diff
@@ -227,10 +232,19 @@ vector2 :: (G.Vector v a, G.Vector v b, ToJSON a, ToJSON b) =>
...
 vector2 name1 name2 v1 v2 = toJSON $ zipWith val (G.toList v1) (G.toList v2) where
     val i j = object
-        [ name1 .= i
-        , name2 .= j
+        [ toKey name1 .= i
+        , toKey name2 .= j
         ]
 
+#if MIN_VERSION_aeson(2,0,0)
+toKey :: T.Text -> Key.Key
```
RyanGlScott, line 240: Out of curiosity, is there a library out there that backports `Key`/`KeyMap`/etc. so that one can do this without CPP? I'm not necessarily wanting to use such a library in `criterion`, but it would be handy in other applications of mine where incurring extra dependencies isn't as big of an issue.

RyanGlScott (approved):
Thanks for the prompt fix!

Criterion/Report.hs
phadej: I'd rather recommend to bump the lower bump on `aeson` if you want to avoid CPP.

We can make a new major version of `aeson-compat` to try to shim this, but I think it will be wasted work, as the need is transient.

PR #248, opened 2021-12-06
escape: remove too agressive JSON escaping

### LINKED ISSUE #235: Presence of double quotes in benchmark name produces broken HTML report
Similarly to #202 and #204, calling `bench` with a string containing double quotes leads to a HTML report that produces a JavaScript error in the browser console. Tested with criterion 1.5.9.0.

For example, generating a HTML report with this variant of the code from #202 illustrates the issue:
```hs
module Main where

import Criterion
import Criterion.Main

main :: IO ()
main = defaultMain
    [ env (return ()) $
       \ ~() -> bgroup "\"oops\"" [bench "dummy" $ nf id ()]
    ]
```

considerate:
This PR removes the overzealous "for good measure" escape of the JSON data that caused issues when using quotes in the benchmark names.

Closes #235

### DIFF
--- a/Criterion/Report.hs
+++ b/Criterion/Report.hs
@@ -119,8 +119,14 @@ report reports = do
 -- (should only affect old versions of IE).
 --
 -- The following characters are replaced with their unicode escape sequnces
--- (\uXXXX) <, >, &, +, \0, \n, \r, ' (single quote), /, \, \x2028 (line
--- separator) and \x2029 (paragraph separator)
+-- (\uXXXX):
+-- <, >, &, +, \x2028 (line separator), and \x2029 (paragraph
+-- separator)
+--
+-- Other characters are such as \\ (backslash) and \n (newline) are not escaped
+-- here as the JSON serializer @encodeToLazyText@ already escapes them when
+-- they occur inside JSON strings and they cause no issues with respect to HTML
+-- safety when used outside of strings in the JSON-encoded payload.
 escapeJSON :: Char -> TL.Text
 escapeJSON '<'      = "\\u003c" -- ban closing of the script tag by making </ impossible
 escapeJSON '>'      = "\\u003e" -- encode tags with unicode escape sequences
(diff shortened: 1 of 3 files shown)


examples/Quotes.hs
```diff
@@ -0,0 +1,10 @@
+module Main where
+
+import Criterion
+import Criterion.Main
+
+main :: IO ()
+main = defaultMain
+    [ env (return ()) $
+       \ ~() -> bgroup "\"oops\"" [bench "dummy" $ nf id ()]
```
RyanGlScott, line 9: Can you also add a benchmark that uses single quotes instead of double quotes? That way, we can test both #202 and #235 simultaneously.

examples/criterion-examples.cabal
```diff
@@ -112,6 +112,17 @@ executable extensible-cli
...
     optparse-applicative
 
+executable quotes
+  main-is: Quotes.hs
+
+  default-language: Haskell2010
+  ghc-options: -Wall -rtsopts
+  build-depends:
+    base == 4.*,
+    base-compat-batteries,
```
RyanGlScott, line 122: This dependency isn't being used, so remove it.

examples/criterion-examples.cabal
```diff
@@ -112,6 +112,17 @@ executable extensible-cli
...
+    criterion,
+    optparse-applicative
```
RyanGlScott, line 124: This dependency isn't being used either. Remove it.

Criterion/Report.hs
```diff
@@ -129,11 +129,6 @@ escapeJSON '\x2029' = "\\u2029" -- paragraph separator
 escapeJSON '&'      = "\\u0026" -- avoid HTML entities
 escapeJSON '+'      = "\\u002b" -- + can be used in UTF-7 escape sequences
 escapeJSON '\0'     = "\\u0000" -- make null characters explicit
-escapeJSON '\n'     = "\\u000a" -- for good measure also escape newlines
```
RyanGlScott, line 132: The Haddocks for `escapeJSON` above mention that `\n`, `\\` and friends are escaped, which is no longer true. Please update the Haddocks, and perhaps mention why `\\` _isn't_ escaped.

RyanGlScott (commented):
Thanks so much, @considerate!

examples/Quotes.hs
considerate: Thank you for the suggest, I have now added a similar benchmark that uses single quotes and tested that the generated JSON and HTML looks correct.

RyanGlScott (approved):
Thanks!

PR #250, opened 2021-12-08
Expand documentation of `escapeJSON`

considerate:
As a follow up to #248 I added some documentation about why \0 is escaped and what to do if the escaped string is to be used in an HTML attribute.

In criterion the JSON-encoded string is used in a `<script>` tag as text (which is indicated in the comment) but I think it would be improper to leave the attribute case unclear, especially since the previous version tried to escape for this use case as well.

### DIFF
--- a/Criterion/Report.hs
+++ b/Criterion/Report.hs
@@ -113,20 +113,26 @@ report reports = do
 -- the <script> tag from within the JSON data is disallowed, i.e, the character
 -- sequence "</" is made impossible.
 --
--- Moreover, single quotes are escaped such that embedding JSON into HTML
--- attributes quoted with single quotes is safe, & is escaped to avoid HTML
--- character references (&<code>;) and + is escaped to avoid UTF-7 attacks
--- (should only affect old versions of IE).
+-- Moreover, & is escaped to avoid HTML character references (&<code>;), + is
+-- escaped to avoid UTF-7 attacks (should only affect old versions of IE), and
+-- \0 is escaped to allow it to be represented in JSON, as the NUL character is
+-- disallowed in JSON but valid in Haskell characters.
 --
--- The following characters are replaced with their unicode escape sequnces
+-- The following characters are replaced with their unicode escape sequences
 -- (\uXXXX):
--- <, >, &, +, \x2028 (line separator), and \x2029 (paragraph
--- separator)
+-- <, >, &, +, \x2028 (line separator), \x2029 (paragraph separator), and \0
+-- (null terminator)
 --
 -- Other characters are such as \\ (backslash) and \n (newline) are not escaped
--- here as the JSON serializer @encodeToLazyText@ already escapes them when
--- they occur inside JSON strings and they cause no issues with respect to HTML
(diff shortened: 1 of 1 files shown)


Criterion/Report.hs
```diff
@@ -113,20 +113,26 @@ report reports = do
...
--- Moreover, single quotes are escaped such that embedding JSON into HTML
--- attributes quoted with single quotes is safe, & is escaped to avoid HTML
--- character references (&<code>;) and + is escaped to avoid UTF-7 attacks
--- (should only affect old versions of IE).
+-- Moreover, & is escaped to avoid HTML character references (&<code>;), + is
+-- escaped to avoid UTF-7 attacks (should only affect old versions of IE), and
+-- \0 is escaped to allow it to be represented in JSON as the NUL character is
+-- disallowed in JSON but valid in Haskell characters.
 --
 -- The following characters are replaced with their unicode escape sequnces
```
RyanGlScott, line 121:
```suggestion
-- The following characters are replaced with their unicode escape sequences
```

Criterion/Report.hs
```diff
@@ -113,20 +113,26 @@ report reports = do
...
+-- (null terminator)
 --
 -- Other characters are such as \\ (backslash) and \n (newline) are not escaped
--- here as the JSON serializer @encodeToLazyText@ already escapes them when
--- they occur inside JSON strings and they cause no issues with respect to HTML
+-- as the JSON serializer @encodeToLazyText@ already escapes them when they
+-- occur inside JSON strings and they cause no issues with respect to HTML
 -- safety when used outside of strings in the JSON-encoded payload.
+--
+-- If the resulting JSON-encoded Text is embedded in an HTML attribute extra
```
RyanGlScott, line 131:
```suggestion
-- If the resulting JSON-encoded Text is embedded in an HTML attribute, extra
```

Criterion/Report.hs
```diff
@@ -113,20 +113,26 @@ report reports = do
 -- the <script> tag from within the JSON data is disallowed, i.e, the character
 -- sequence "</" is made impossible.
 --
--- Moreover, single quotes are escaped such that embedding JSON into HTML
--- attributes quoted with single quotes is safe, & is escaped to avoid HTML
--- character references (&<code>;) and + is escaped to avoid UTF-7 attacks
--- (should only affect old versions of IE).
+-- Moreover, & is escaped to avoid HTML character references (&<code>;), + is
+-- escaped to avoid UTF-7 attacks (should only affect old versions of IE), and
+-- \0 is escaped to allow it to be represented in JSON as the NUL character is
```
RyanGlScott, line 118:
```suggestion
-- \0 is escaped to allow it to be represented in JSON, as the NUL character is
```

RyanGlScott (commented):
Thanks! I have some minor grammatical and spelling suggestions, but otherwise this LGTM.

considerate:
> Thanks! I have some minor grammatical and spelling suggestions, but otherwise this LGTM.

Thank you! I updated the documentation according to your comments.

RyanGlScott (approved):

PR #251, opened 2021-12-17
Add xhtml template

414owen:
This template passes (https://validator.w3.org/) under XHTML 1.0 Strict, and renders correctly.

It's basically identical to the `html` template, with a few tags changed to satisfy the validator.

## To use

```
./prog --template xhtml --output benchmark.xhtml
```

## Sample

https://owen.cafe/advent/2021/

(benchmarking advent of code 2021 solutions)


414owen:
Closing until I figure out better polyglot markup support.

Issue #218, opened 2019-12-12
Feature request: allow limiting very long-running tests


jberryman:
I would like a flag that lets me say "never spend more than X seconds on a benchmark; if the time limit is exceeded compile statistics with runs you've done".

I assumed from help that this was what `-L` did, but I was wrong (I think those docs could be improved too; `-L` seems to run a benchmark for some fixed amount of time which we wouldn't normally call a "time limit"... I wonder if it was originally meant to do what I thought it did...).

jberryman:
maybe dupe of https://github.com/bos/criterion/issues/210

daberg:
Hello there, I have the same problem as @jberryman. Are there any news about this enhancement? Any possible workarounds? Thanks.

RyanGlScott:
I'm not actively working on this, but I would welcome a PR that implemented it.

Issue #210, opened 2019-02-01
How can I do shorter benchmarks?


chrissound:
```
module Main where

import Criterion.Types
import Criterion.Main

myConfig :: Config
myConfig = defaultConfig {
              resamples = 1
           , timeLimit = 0.01
           }

main :: IO ()
main = do
  let f = (\x -> bench (show x)  $ nf xyz x)
  defaultMainWith myConfig [
    bgroup "fib" [
                  env (pure 400000) f
                , env (pure 100) f
                , env (pure 200) f
                ]
    ]

xyz :: Int -> [Double]
xyz 0 = []
xyz _ = [fromIntegral 1000]

```

I was hoping I could quickly benchmark an example like this - but it seems to be taking around 4 seconds - to do numerous(?) runs. 

Have I misunderstood the `resamples` / `timeLimit` values?

RyanGlScott:
In order to ensure that `criterion` generates enough data for us to have some trust in the raw measurements, `criterion` sets a minimum time threshold that a benchmark must run for:

https://github.com/bos/criterion/blob/642bea6d15b9a1737c82ffd051f663edcb89d280/criterion-measurement/src/Criterion/Measurement.hs#L240-L250

This is where the `threshold` is used in the inner loop of benchmarking:

https://github.com/bos/criterion/blob/642bea6d15b9a1737c82ffd051f663edcb89d280/criterion-measurement/src/Criterion/Measurement.hs#L300-L315

If you do the math here, it will likely add up to the time you're seeing.

Admittedly, the Haddocks for `timeLimit` don't make mention of this fact, although they probably should.

jberryman:
I have some benchmarks where a single iteration takes ~5min. Is there any alternative for me other than rewriting them outside of criterion? I essentially want to run them just once and have the time reported.

PR #255, opened 2022-01-12
Add runBenchmarkWith, general benchmark runner

### LINKED ISSUE #254: Add a variant of runBenchmark which also takes iteration range.
Currently there is:

```haskell
runBenchmark :: Benchmarkable -> Double -> IO (Vector Measured, Double)
```

> Run a single benchmark, and return measurements collected while executing it, along with the amount of time the measurement process took.

> Double argument is: Lower bound on how long the benchmarking process should take.  In practice, this time limit may be exceeded in order to generate enough data to perform meaningful statistical analyses.

I'd like to have a variant where we can specify the optional lower and upper bounds for iteration count (run at least 10 times, no need to run over 100 times).

Lower bound would be great to ensure more statistical significance. Upper bound would allow to avoid running individual benchmark for tool long.

I think that then we'd need both lower and upper bound for time as well (current is actually more like soft upper bound, but as there are no end criteria it's also a lower bound).

I think it would best to put these four options into a record, to avoid double-blindness.
In case when no bounds is specified, we could either fail or use some default time bound.

phadej:
Refactor runBenchmark to use it.
Resolves #254

---

I opted out to generating some data-type with configuration etc. The logic is just easier to write as code directly.

### DIFF
--- a/criterion-measurement/src/Criterion/Measurement.hs
+++ b/criterion-measurement/src/Criterion/Measurement.hs
@@ -27,6 +27,7 @@ module Criterion.Measurement
     , secs
     , measure
     , runBenchmark
+    , runBenchmarkWith
     , runBenchmarkable
     , runBenchmarkable_
     , measured
@@ -39,7 +40,7 @@ import Control.DeepSeq (NFData(rnf))
 import Control.Exception (finally,evaluate)
 import Data.Data (Data, Typeable)
 import Data.Int (Int64)
-import Data.List (unfoldr)
+import Data.List.NonEmpty (NonEmpty (..))
 import Data.Word (Word64)
 import GHC.Generics (Generic)
 #if MIN_VERSION_base(4,10,0)
@@ -58,6 +59,7 @@ import Text.Printf (printf)
 import qualified Control.Exception as Exc
 import qualified Data.Vector as V
 import qualified GHC.Stats as Stats
+import qualified Data.List.NonEmpty as NE
 
 #if !(MIN_VERSION_base(4,7,0))
 foreign import ccall "performGC" performMinorGC :: IO ()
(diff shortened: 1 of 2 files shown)


phadej:
I'll add semigroups compat if this is otherwise ok, or when I make tweaks.

criterion-measurement/src/Criterion/Measurement.hs
```diff
@@ -289,41 +291,69 @@ runBenchmark :: Benchmarkable
              -- exceeded in order to generate enough data to perform
              -- meaningful statistical analyses.
              -> IO (V.Vector Measured, Double)
-runBenchmark bm timeLimit = do
+runBenchmark bm timeLimit = runBenchmarkWith endAt 0 bm where
+  endAt :: Int -> Int64 -> Double -> NonEmpty Measured -> Double -> Maybe (Int64, Double)
```
RyanGlScott, line 295: Since this is the default behavior for `runBenchmarkWith`'s termination check, I'd favor factoring this out into a top-level definition so that a user can use it if they wish to tweak it slightly. Since this definition closes over `timeLimit`, it might make things more clear to turn this into a newtype:

```hs
newtype TerminationCheck s = TerminationCheck
  { getTerminationCheck :: Int -> Int64 -> Double -> NonEmpty Measured -> s -> Maybe (Int64, s) }

defaultTerminationCheck :: Double -> TerminationCheck s
defaultTerminationCheck timeLimit = TerminationCheck $ \count iters delta ms prev -> do ...
```

Then you can mention in the Haddocks for `runBenchmark` that `runBenchmark bm timeLimit = runBenchmarkWith (defaultTerminationCheck timeLimit) 0 bm`.

criterion-measurement/src/Criterion/Measurement.hs
```diff
@@ -289,41 +291,69 @@ runBenchmark :: Benchmarkable
...
+-- * current run count (starting from zero)
+-- * current iteration count (starting from one)
+-- * amount of time the measurement process took so far.
+-- * results so far (in reverse order, latest first in the list).
+-- * user defined state
+--
+-- and should return 'Nothing' for run to terminate
+-- or @'Just' (nextIters, nextState)@ for run to continue with new iteration count and state.
+--
+runBenchmarkWith :: forall s. (Int -> Int64 -> Double -> NonEmpty Measured -> s -> Maybe (Int64, s))
```
RyanGlScott, line 334: Speaking of which, what _is_ the best way to tweak the default behavior? I imagine that a common use case would be to use `runBenchmarkWith` to impose a lower maximum on the number of iterations being run, as per #210/#218. Is this straightforward to accomplish with this design?

criterion-measurement/src/Criterion/Measurement.hs
```diff
@@ -289,41 +291,69 @@ runBenchmark :: Benchmarkable
...
+-- The first argument is termination check, it takes as arguments:
+--
+-- * current run count (starting from zero)
+-- * current iteration count (starting from one)
+-- * amount of time the measurement process took so far.
+-- * results so far (in reverse order, latest first in the list).
+-- * user defined state
+--
+-- and should return 'Nothing' for run to terminate
+-- or @'Just' (nextIters, nextState)@ for run to continue with new iteration count and state.
```
RyanGlScott, line 332: This needs a loud disclaimer that you should be careful when picking a termination check, as its implementation could affect the statistical quality of the data.

criterion-measurement/src/Criterion/Measurement.hs
```diff
@@ -289,41 +291,69 @@ runBenchmark :: Benchmarkable
...
+
+  -- Multiply by 1.05, but always increase
+  nextIters :: Int64 -> Int64
+  nextIters n = max (n + 1) (div (105 * n) 100)
+
+-- | Run benchmark with provided dynamic termination condition.
+--
+-- The first argument is termination check, it takes as arguments:
+--
+-- * current run count (starting from zero)
```
RyanGlScott, line 325: If `TerminationCheck` were factored out into a newtype, these Haddocks could be moved to the newtype instead.

criterion-measurement/src/Criterion/Measurement.hs
phadej: I'm not so sure about. That function needs to be factored so its pieces could be  usable, but I don't know how we could factor it. There are at least two parts: updating the state and deciding termination, but that consist of three checks already.

criterion-measurement/src/Criterion/Measurement.hs
phadej: I don't see an other way then copying the source and tweaking it.

criterion-measurement/src/Criterion/Measurement.hs
phadej: [`analyseSample`](https://hackage.haskell.org/package/criterion-1.5.12.0/docs/Criterion-Analysis.html#v:analyseSample) doesn't have any pre condiotions specified. Looking at the code, it apperently needs at least 4 samples, then the statistical quality will be estimated (in the confidence interval).


I.e. I don't know how to put it "Termination check condition should be picked so the data generated is suitable for whatever further analysis is performed on it". Isn't that obvious?

phadej:
I changed the type of function. It fits better the usage where some analysis is done already on each step (in ~bayesian style), so you won't end doing it twice on the last sample (and without access to the state).

criterion-measurement/src/Criterion/Measurement.hs
RyanGlScott: I think I explained myself poorly in https://github.com/haskell/criterion/pull/255#discussion_r783494620, so let me try again. I agree that because termination checks will need to perform arbitrary computation, there's not a good way in general to configure everything that a termination check might want to do. That being said, I still think there is value in exposing the default termination check to users. This is because in #218, a user wishes to do what `runBenchmark` does plus an additional constraint on the number of iterations performed. In terms of code, this could look something like this:

```hs
longLivedComputationCheck :: TerminationCheck r s
longLivedComputationCheck = TerminationCheck $ \... -> do
  getTerminationCheck defaultTerminationCheck ...
  if <time exceeds certain threshold>
    then Left ...
    else Right ...
```

For some use cases, there might not be a better alternative than having to copy-paste the source code of `defaultTerminationCheck` and tweaking it, but I don't want that to always be the case.

criterion-measurement/src/Criterion/Measurement.hs
RyanGlScott: To be clear, I'm not looking for a super-detailed analysis of _how_ to make the data statistically significant. I'm just looking for a generic warning saying that you should exercise caution when using this function, as it has the potential to distort the meaningfulness of the data if implemented haphazardly.

phadej:
I don't need this functionality anymore.

Issue #256, opened 2022-01-17
[FR] support benchmarking max memory usage


NightMachinary:
Migrated from https://github.com/Gabriel439/bench/issues/45#issuecomment-1014904269 .

RyanGlScott:
A place to start would be to record the [`max_mem_in_use_bytes`](https://hackage.haskell.org/package/base-4.16.0.0/docs/GHC-Stats.html#v:max_mem_in_use_bytes) from the `RTSStats` in the [`getStatistics`](https://github.com/haskell/criterion/blob/3f6a32b01dbadb5c0954aaebaf4fef923e7ac7ec/criterion-measurement/src/Criterion/Measurement.hs#L121) function. Then you'd need to update the maximum in the [`addResults`](https://github.com/haskell/criterion/blob/3f6a32b01dbadb5c0954aaebaf4fef923e7ac7ec/criterion-measurement/src/Criterion/Measurement.hs#L219) function. (If we start recording maximum values, however, the name `addResults` isn't really accurate anymore, since you wouldn't add two maximum values together.)

RyanGlScott:
Thanks for working on this, @tbidne! Please do open a PR, as it would be more convenient to review the patch using the GitHub review UI. From a quick glance, your patch looks like what I was imagining it would be.

RyanGlScott:
Fixed in #260.

PR #260, opened 2022-03-02
Add peak bytes allocated field to Measured

tbidne:
Adds the "peak mb allocated" to the json output for #256.

I wasn't sure which "max mb" stat we wanted to take here, so I took the final `endPostGC` one. In my testing `endPostGC`, `endStatsPreGC`, and `startStats` were all the same, in any case. 
```haskell
  measAllocated         = diff endPostGC gcStatsBytesAllocated
, measPeakMbAllocated   = gcStatsPeakMegabytesAllocated endPostGC
```

Example:
```json5
{
  "reportMeasured": [
    [
      4.930255000090256e-3,
      1.1697070000000007e-3,
      11361858,
      1,
      1236872,
      3,                /* peakMbAllocated */
      1,
      189848,
      4.644875e-3,
      8.848910000000001e-4,
      2.9445099999999983e-4,
      2.93705e-4
    ]
  ],
  ...
  "reportKeys": [
    "time",
    "cpuTime",
    "cycles",
    "iters",
    "allocated",
    "peakMbAllocated",
    "numGcs",
    "bytesCopied",
    "mutatorWallSeconds",
    "mutatorCpuSeconds",
    "gcWallSeconds",
    "gcCpuSeconds"
  ]
}
```

### DIFF
--- a/criterion-measurement/src/Criterion/Measurement/Types.hs
+++ b/criterion-measurement/src/Criterion/Measurement/Types.hs
@@ -125,6 +125,8 @@ data Measured = Measured {
 
     , measAllocated          :: !Int64
       -- ^ __(GC)__ Number of bytes allocated.  Access using 'fromInt'.
+    , measPeakMbAllocated    :: !Int64
+      -- ^ __(GC)__ Max number of megabytes allocated.  Access using 'fromInt'.
     , measNumGcs             :: !Int64
       -- ^ __(GC)__ Number of garbage collections performed.  Access
       -- using 'fromInt'.
@@ -148,11 +150,11 @@ data Measured = Measured {
 
 instance FromJSON Measured where
     parseJSON v = do
-      (a,b,c,d,e,f,g,h,i,j,k) <- parseJSON v
+      (a,b,c,d,e,f,g,h,i,j,k,l) <- parseJSON v
       -- The first four fields are not subject to the encoding policy:
       return $ Measured a b c d
-                       (int e) (int f) (int g)
-                       (db h) (db i) (db j) (db k)
+                       (int e) (int f) (int g) (int h)
+                       (db i) (db j) (db k) (db l)
       where int = toInt; db = toDouble
 
 -- Here we treat the numeric fields as `Maybe Int64` and `Maybe Double`
@@ -161,7 +163,7 @@ instance FromJSON Measured where
 instance ToJSON Measured where
     toJSON Measured{..} = toJSON
       (measTime, measCpuTime, measCycles, measIters,
-       i measAllocated, i measNumGcs, i measBytesCopied,
+       i measAllocated, i measPeakMbAllocated, i measNumGcs, i measBytesCopied,
        d measMutatorWallSeconds, d measMutatorCpuSeconds,
        d measGcWallSeconds, d measGcCpuSeconds)
       where i = fromInt; d = fromDouble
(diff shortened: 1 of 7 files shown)


criterion-measurement/src/Criterion/Measurement/Types.hs
```diff
@@ -125,6 +125,8 @@ data Measured = Measured {
 
     , measAllocated          :: !Int64
       -- ^ __(GC)__ Number of bytes allocated.  Access using 'fromInt'.
+    , measPeakMbAllocated    :: !Int64
```
RyanGlScott, line 128: It would definitely be worth noting the addition of this new field to the `changelog`. Moreover, it is also worth noting that the behavior of `Measured`'s `Binary` and `ToJSON` instances have changed, since it's possible that there are user programs that deserialize `Measured` values directly from binary or JSON. (I don't think that rises to the level of needing a major version bump or anything, but it is worth at least a mention.)

criterion-measurement/src/Criterion/Measurement/Types.hs
```diff
@@ -186,6 +188,8 @@ measureAccessors_ = [
                             "loop iterations"))
   , ("allocated",          (fmap fromIntegral . fromInt . measAllocated,
                             "(+RTS -T) bytes allocated"))
+  , ("peakMbAllocated",    (fmap fromIntegral . fromInt . measPeakMbAllocated,
+                            "(+RTS -T) peak bytes allocated"))
```
RyanGlScott, line 192: Should this read megabytes instead?

```suggestion
                            "(+RTS -T) peak megabytes allocated"))
```

RyanGlScott (commented):
Thanks, @tbidne! This looks fantastic. And I agree that your choice of `endPostGC` makes sense.

A couple of minor suggestions inline.

criterion-measurement/src/Criterion/Measurement/Types.hs
tbidne: Yep, good catch!

criterion-measurement/src/Criterion/Measurement/Types.hs
tbidne: Good point. Actually, I think this probably _does_ require a major version change for `criterion-measurement` and `criterion` because both export `Measured`, which has the new `measPeakMbAllocated` field.

RyanGlScott: Ah, I had forgotten that `Measured` is exported. In that case, I agree that major version bumps are in order.

tbidne:
Okay, comments addressed. Thanks for the quick feedback!

tbidne:
@NightMachinary would this address your request?

RyanGlScott (approved):

RyanGlScott:
Excellent work—thanks again for the patch. Would you prefer to have a new Hackage release with these changes soon? If not, I'll get around to it eventually when I next upload a round of packages to Hackage.

tbidne:
@RyanGlScott Thanks for merging! No rush on my account. I understand the major version bump may provoke a desire to include more such changes to avoid a lot of "churn", and if so that's fine.

I do wonder if it would be desirable to have a "prettier" way to display consumed memory than the json output, for both allocated memory and the new peak memory (perhaps @NightMachinary would agree); Scanning the json output for each individual run -- AFAIK the only way to see such data -- is not as friendly as the consolidated data in the html and csv files. In other words, `criterion` seems more geared towards displaying timing data than memory usage. We could imagine, for instance, text output that shows the average peak memory across all runs, in addition to the absolute max.

But that probably requires a separate issue, and I'm unable to judge how well that fits into `criterion`'s intended usage. I am happy to work on this, but I'd need some guidance.

Anyway, thanks again for the quick feedback and merge :slightly_smiling_face:.

PR #263, opened 2022-05-22
Benchmarking linear functions

yaitskov:


RyanGlScott:
Thanks for the pull request.

Before I review this PR, I think we should settle on a name for this function. I'm not a fan of the name "`lf`" for a couple of reasons:

1. It breaks the conventions somewhat. The `nf` function, for instance, stands for "normal form", but `lf` doesn't stand for "linear form".
2. Although this PR doesn't add them, we might conceivably want linear versions of the `whnf`, `nfAppIO`, and `whnfAppIO` functions in the future. I'd prefer to have a naming scheme that gracefully extend to these functions as well.

With that in mind, some possibilities that come to mind:

1. Name it "`nfLinear`". Then other variants would be named `whnfLinear`, `nfAppIOLinear`, etc.
2. Define `nf`, but in a separate module with a `.Linear` suffix, similarly to how `linear-base`'s naming conventions work.

Do either of these options work for you, @yaitskov/

yaitskov:
@RyanGlScott, thanks - the review was fast. I will rename function accordingly.
I thought about weak headed normal forms in the context of linear functions.
IMHO whnf is not needed for a linear function, 
because GHC type checker guaranties - all fields are consumed exactly once.
e.g. the following snippet won't compile:

```
import Prelude.Linear as Linear

data McMeal = McMeal { mcDrink :: String, mcBurger :: String } deriving (Linear.Show)

eatCalories :: McMeal %1-> String
eatCalories mc = mcDrink mc -- ++ mcBurger mc

```

```
    • Couldn't match type ‘'Many’ with ‘'One’
        arising from multiplicity of ‘mc’
    • In an equation for ‘eatCalories’: eatCalories mc = mcDrink mc
   |
91 | eatCalories mc = mcDrink mc -- ++ mcBurger mc

```

So `nfLinear = whnfLinear` isn't it?

RyanGlScott:
> IMHO whnf is not needed for a linear function,
> because GHC type checker guaranties - all fields are consumed exactly once.

I don't think so. Consuming a linear data type will imply that its _linear_ fields are consumed, but it doesn't say anything about its _non-linear_ fields. For example, suppose you had written this:

```hs
data McMeal where
  McMeal :: String %1 -> String -> McMeal

eatCalories :: McMeal %1-> String
eatCalories (McMeal drink _burger) = drink
```

Here, `McMeal`'s first field is linear, but not its second. Here, WHNF is definitely not equal to NF, as demonstrated by the fact that we evaluated `McMeal` to WHNF but did not evaluate the second field, even though the `McMeal` value was consumed linearly.

RyanGlScott:
Thanks, @yaitskov.

Another design issue, which is worth bringing up: do we even want to have separate `nf` and `nfLinear` functions in the first place? That is, what if we instead made `nf` multiplicity polymorphic on sufficiently new GHCs? Something like:

```hs
#if MIN_VERSION_base(4,15,0)
nf :: forall a b p. NFData b => (a %p -> b) -> a -> Benchmarkable
#else
nf :: forall a b. NFData b => (a -> b) -> a -> Benchmarkable
#endif
```

I've tested this idea locally, and it appears to work. One nice thing about this change is that none of the implementations of `nf` or related functions would need to change, only their types. Moreover, code using `nf` will continue to work, and anyone who wants to use `criterion` in a linear setting can use the same functions to do so.

What do you think of this plan? There's an appealing simplicity to this idea, but I'd want to make sure that it would work for your use case.

Bodigrim:
Linear functions can be benchmarked by existing `nf` just fine: `nf (\x -> f x) a` should work for linear `f`.

PR #264, opened 2022-08-04
Make nf' and whnf' robust against eta-expansion changes

mpickering:
As seen in ghc#21948 (https://gitlab.haskell.org/ghc/ghc/-/issues/21948), whether
or not `go` gets eta-expanded crucially determines whether `nf'` is a
function which computes an `IO ()` action or a function which computes a
result of type `(# RealWorld State#, () #)`.

In the former case, the call to `seq` in `runBenchmark` will force the
computation of the `IO ()` action, which can be computed without
dependency on the State# token, and therefore before any IO takes place. This
defeats the point of the function which is supposed to run the loop
inside the IO action (so that measure can correct measure the runtime
stats).

The fix is to make both branches of the loop depend on the state token so
that the loop can't be evaluated before the whole IO action is executed.

### DIFF
--- a/criterion-measurement/src/Criterion/Measurement/Types/Internal.hs
+++ b/criterion-measurement/src/Criterion/Measurement/Types/Internal.hs
@@ -20,6 +20,7 @@
 module Criterion.Measurement.Types.Internal (fakeEnvironment, nf', whnf') where
 
 import Data.Int (Int64)
+import Control.Exception
 
 -- | A dummy environment that is passed to functions that create benchmarks
 -- from environments when no concrete environment is available.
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Thanks for the fix, @mpickering!

I'd like to make sure that we document why we are using `evaluate` to ensure that we don't accidentally revert this change if this code becomes refactored in the future. Can you leave a comment somewhere in [here](https://github.com/haskell/criterion/blob/3f6a32b01dbadb5c0954aaebaf4fef923e7ac7ec/criterion-measurement/src/Criterion/Measurement/Types/Internal.hs#L34-L48) that explains the rationale for using `evaluate` over `seq`? Thanks!

mpickering:
I added a comment @RyanGlScott

RyanGlScott (approved):

PR #265, opened 2022-08-07
Remove entries from www/ directory from the linguist statistics

lyokha:
This should update the project's language bar and tag it as a Haskell project (currently, Criterion is tagged as an HTML project).

This is what I mean (after applying this PR):

criterion-linguist-bar


RyanGlScott (approved):
Thanks!

PR #266, opened 2022-08-19
Remove dependency on deprecated library ansi-wl-pprint

newhoggy:


RyanGlScott:
What is the difference between this PR and #242? In light of [this comment](https://github.com/haskell/criterion/pull/242#issuecomment-884796365) on #242, has `optparse-applicative` switched to `prettyprinter` upstream, or is it still using (and re-exporting) combinators from `ansi-wl-pprint`?

newhoggy:
Sorry, it's been so long I forgot I've created this PR before.  I'll re-integrate this with the other.

PR #276, opened 2023-05-23
Require `optparse-applicative-0.18.*` as the minimum

RyanGlScott:
Also add explicit dependencies on `prettyprinter` and `prettyprinter-ansi-terminal`. With this change, we have expunged any traces of `ansi-wl-pprint` from `criterion`.

This builds on top of #275.


RyanGlScott:
Come to think of it, this doesn't really require a major version bump, as it only changes the internals of `criterion`.

PR #282, opened 2024-05-12
Support non-Darwin, non-Linux, aarch64 platforms

depressed-pho:
We can read the `cntvct_el0` register to observe the CPU clock. It is a single instruction and has barely any overhead.


RyanGlScott (approved):
Thanks!

RyanGlScott:
I've uploaded [`criterion-measurement-0.2.2.0`](https://hackage.haskell.org/package/criterion-measurement-0.2.2.0) to Hackage with these changes.

PR #288, opened 2026-04-01
Fix typos

kianmeng:
Found via `codespell -L te,ot,ue`

### DIFF
--- a/criterion-measurement/src/Criterion/Measurement/Types.hs
+++ b/criterion-measurement/src/Criterion/Measurement/Types.hs
@@ -503,7 +503,7 @@ envWithCleanup = Environment
 -- The environment is evaluated to normal form before the benchmark is run.
 --
 -- When using 'whnf', 'whnfIO', etc. Criterion creates a 'Benchmarkable'
--- whichs runs a batch of @N@ repeat runs of that expressions. Criterion may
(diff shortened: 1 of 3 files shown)


criterion-measurement/src/Criterion/Measurement/Types.hs
```diff
@@ -503,7 +503,7 @@ envWithCleanup = Environment
 -- The environment is evaluated to normal form before the benchmark is run.
 --
 -- When using 'whnf', 'whnfIO', etc. Criterion creates a 'Benchmarkable'
--- whichs runs a batch of @N@ repeat runs of that expressions. Criterion may
+-- which's runs a batch of @N@ repeat runs of that expressions. Criterion may
```
RyanGlScott: I don't think this is quite right either. I would phrase this as:


```suggestion
-- which runs a batch of @N@ repeat runs of that expressions. Criterion may
```

RyanGlScott (commented):
Thanks! (The CI failures are unrelated---see https://github.com/haskell/criterion/pull/290.)

RyanGlScott (approved):
Thanks!

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #290, opened 2026-04-01
tests/Cleanup: prevent parallel test execute with tasty >= 1.5.4

sternenseemann:
tasty 1.5.4 automatically parallelizes test cases in a testGroup unless told not to when built with -threaded (see
https://hackage.haskell.org/package/tasty-1.5.4/changelog). This differs from the old behavior which required +RTS -N.

Since the cleanup test suite uses shared files for all test cases, it can't be parallelized at the moment.


RyanGlScott (approved):
Thanks again!

PR #291, opened 2026-07-01
cycles.c: add powerpc64* support

pkubaj:
Currently build fails on FreeBSD/powerpc64{le,} without that.

### DIFF
--- a/criterion-measurement/cbits/cycles.c
+++ b/criterion-measurement/cbits/cycles.c
@@ -18,6 +18,19 @@ StgWord64 criterion_rdtsc(void)
   return ret;
 }
 
+#elif powerpc64le_HOST_ARCH || powerpc64_HOST_ARCH
+
+StgWord64 criterion_rdtsc(void)
+{
+  // Read the PowerPC Time Base: a monotonically-increasing 64-bit counter
+  // that is the architectural analogue of x86's TSC. SPR 268 returns the
+  // full 64-bit Time Base on 64-bit PowerPC (this is what the deprecated
+  // `mftb` mnemonic expanded to).
+  StgWord64 ret;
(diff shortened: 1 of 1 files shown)


criterion-measurement/cbits/cycles.c
```diff
@@ -18,6 +18,15 @@ StgWord64 criterion_rdtsc(void)
   return ret;
 }
 
+#elif powerpc64le_HOST_ARCH || powerpc64_HOST_ARCH
+
+StgWord64 criterion_rdtsc(void)
+{
+  StgWord64 ret;
+  __asm__ __volatile__ ("mfspr %0, 268" : "=r"(ret));
```
RyanGlScott, line 30: Can you leave a comment here saying what the significance of the number `268` is? I'm not a PPC expert, but my understanding is that `268` is a special-purpose register of some kind.

RyanGlScott (commented):
Great, thank you for the pull request! One minor suggestion, but otherwise I'm happy to land this.

criterion-measurement/cbits/cycles.c
pkubaj: Done.

RyanGlScott (approved):
Thank you!

RyanGlScott:
I've uploaded [`criterion-measurement-0.2.5.0`](https://hackage.haskell.org/package/criterion-measurement-0.2.5.0) to Hackage with these changes.

pkubaj:
The changelog says:
* Support building on PowerPC.

But it's not really true. That instruction only works on 64-bit PowerPC, won't work on 32-bit. This is how glibc does it:
https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/powerpc/sys/platform/ppc.h#l33

RyanGlScott:
Sorry, my mistake!

